PR #1, opened 2017-07-30
add opencv_videoio to cffi dependencies

Adimote:
add `opencv_videoio` to cffi build dependencies as my machine cannot find symbol `_ZN2cv12VideoCaptureC1Ei` without it.


Adimote:
As this fails on a debian machine, it must be an error in my own setup (suspect too new version of opencv, I have 3.x.x installed), I'll cancel this PR

PR #3, opened 2017-07-30
Distance model via quadratic regression

prophile:
This uses the calibration step to fit quadratic regression coefficients on the homography matrix.


sb_vision/calibration/training_example.py
```diff
@@ -1,10 +1,36 @@
...
+    def _load(self):
+        """Load and recognise the single marker from this image."""
+
+        pseudo_camera = FileCamera(self.image_file, focal_length=None)
+        vision = Vision(pseudo_camera, (0.01, 0.01))
+
+        (single_token,) = vision.snapshot()
+
+        # Explicitly clean up vision
+        del vision
```
thomasleese: Why?

sb_vision/calibration/training_example.py
```diff
@@ -1,10 +1,36 @@
...
+
+        self._load()
+
+    def _load(self):
+        """Load and recognise the single marker from this image."""
+
+        pseudo_camera = FileCamera(self.image_file, focal_length=None)
+        vision = Vision(pseudo_camera, (0.01, 0.01))
+
+        (single_token,) = vision.snapshot()
```
thomasleese: This looks a bit odd to me, I would have `single_token, _ = `.

sb_vision/vision.py
```diff
@@ -18,29 +18,14 @@
...
 
-    def __init__(self, camera: CameraBase, token_sizes):
-        """
-        General initialiser.
-
-        In general `token_sizes` must be a mapping of token IDs to (vertical,
-        horizontal) tuples in units of metres, but for convenience it may also
-        be passed as a single (vertical, horizontal) tuple, or a single number.
-        """
+    def __init__(self, camera: CameraBase):
```
thomasleese, line 19: Yeeesssssss types

sb_vision/calibration/training_example.py
```diff
@@ -0,0 +1,41 @@
...
+        self.size = image.size
+        try:
+            (single_token,) = vision.process_image(image)
+        except ValueError:
+            raise RuntimeError(
+                "Cannot detect token in {}".format(self.image_file.name),
+            )
+
+        # Explicitly clean up vision
+        del vision
```
thomasleese: Why?

sb_vision/calibration/cli.py
```diff
@@ -0,0 +1,68 @@
...
+        '-o',
+        '--output',
+        type=Path,
+        help="file to save into",
+    )
+
+    return parser
+
+
+def main(args=sys.argv[1:]):
```
thomasleese: This feels wrong even though there most likely won't be any problems due to it only be called once.

```python
def main(args=None):
    if args is not None:
        args = sys.argv[1:]
```

thomasleese (commented):
Looks good, the last commit is a merge commit which doesn't feel like it should be there.

sb_vision/calibration/training_example.py
prophile: That has different semantics: `(foo,)` unpacks a single-element iterable whereas `foo, _` unpacks a two-element tuple ignoring the second.

I'm not averse to writing this `single_token, =` and dropping the parens if you'd prefer, I tend to include the parens to make it more visually distinct from `single_token =`.

sb_vision/calibration/training_example.py
prophile: `vision` holds some resources which are cleaned up on `__del__`. Worth a comment you think?

sb_vision/vision.py
prophile: they don't work

sb_vision/calibration/cli.py
prophile: 776fa8c

prophile:
The merge commit was due to a conflict in `vision.py` which I thought was worth resolving in this branch.

sb_vision/calibration/training_example.py
thomasleese: Just makes me sad that we have to do this forcibly, I think the comment could be clearer.

prophile: 7c28541

thomasleese (approved):
👍 🎉

sb_vision/tokens.py
```diff
@@ -58,107 +61,103 @@ def _get_pixel_centre(homography_matrix):
...
+    )
+
+    return np.array([x, y, z])
 
 
 class Token:
     """Representation of the detection of one token."""
 
-    def __init__(self, id, size=DEFAULT_TOKEN_SIZE, certainty=0.0):
+    def __init__(self, id, certainty=0.0):
```
PeterJCLaw, line 145: Does the removal of token size mean that we only support a single size of token now? If not, where does the size of the token get taken into account?

prophile: (Discussed IRL: now the responsibility of callers)

PR #7, opened 2017-11-07
Use Pipenv

thomasleese:
This configures the repo to use [Pipenv](https://docs.pipenv.org) as the workflow for development and tests.


RealOrangeOne:
If we're making the change to `Pipenv` (which i'm totally down for), we should probably update all repos to use it too

thomasleese:
@RealOrangeOne That's my plan 😄

README.md
```diff
@@ -1,4 +1,16 @@
...
+## Installation
+
+```bash
+$ pipenv install --dev
+```
+
+## Running Tests
+
+```bash
+$ pipenv run python setup.py test
```
PeterJCLaw, line 14: Seriously? How is this better than being able to `./setup.py $thing`?

PeterJCLaw:
I suggest cherry-picking acffb1f as it's not related to this PR really.

setup.py
```diff
@@ -3,9 +3,9 @@
 
 setup(
     name='sb-vision',
-    version='1.0',
-    description="Vision system for SourceBots",
-    author="SourceBots",
+    version='0.1.0',
```
PeterJCLaw: I'm pretty sure going backwards in version numbers is a terrible idea.

setup.py
```diff
@@ -3,9 +3,9 @@
...
+    description='Vision system for SourceBots',
+    author='SourceBots',
```
PeterJCLaw: I suspect these two lines were double quotes intentionally -- there's a convention developing to use double quotes for strings intended for humans and single quotes for strings for machines.

Pipfile
```diff
@@ -0,0 +1,21 @@
+[[source]]
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
```
PeterJCLaw, line 6: I'm _guessing_ that the `packages` vs `dev-packages` separation here is around which packages get installed when consuming vs developing this package? If so, I literally _don't believe_ that this package has no run-time dependencies.

PeterJCLaw: Note: if this listing is secondary to what's in the `setup.py` file, then I seriously question how Pipenv is meant to work -- duplicating our dependency listing like this is a _bad_ idea.

PeterJCLaw:
Pipenv looks like it has some benefits, though I've not used it so I can't tell whether it's worth it.

I don't like the look of having to re-route all commands via `pipenv run $thing`. I've seen that pattern elsewhere and find it ugly and annoying (often because it breaks tab completions for things).

I also don't like that switching to this _forces_ developers to use pipenv. Previously it was up to the developer to choose how they installed the dependencies (virtualenv or otherwise) by using standard patterns. Pipenv seems to force users to use it by requiring custom file formats, rather than playing nicely with the existing standards.

README.md
thomasleese: It's not equivalent to that, it activates the virtual environment and runs it in one command.

setup.py
thomasleese: I agree, but we haven't made a release yet and generally 1.x is reserved for the first stable release, which this definitely isn't.

Pipfile
thomasleese: These packages are just for setting up the virtual environment. Installing the `sb-vision` package installs the dependencies listed in `setup.py` so we don't need to duplicate them in the `Pipfile`.

thomasleese:
Having to do `pipenv run $thing` is no different to having to do `venv/bin/python`, you can always use `pipenv shell` which is equivalent to `source venv/bin/activate`.

Pipenv is the [standard pattern recommended by the PSF](https://packaging.python.org/new-tutorials/installing-and-using-packages/#installing-pipenv) so I don't think you could say it's forcing users to use custom file formats. Users can still `pip install -e .` the package if they want to.

Pipfile
thomasleese: I concluded that `sb-vision` should live in `dev-packages` simply because it is what users will be developing, but it is slightly odd.

README.md
PeterJCLaw: I realise it's not technically equivalent, though it is functionally the same with more typing and less tab-completion.

setup.py
```diff
@@ -20,13 +20,25 @@
         'Pillow',
         'numpy',
         'scipy',
-        "cffi>=1.4.0",
+        'cffi>=1.4.0',
+        'pytest-runner',
```
PeterJCLaw, line 24: Why do we now have a run-time dependency on `pytest`? This seems like a _bad_ thing.

Pipfile
```diff
@@ -0,0 +1,21 @@
...
+
+[dev-packages]
+
+"e1839a8" = {path = ".", editable = true}
```
PeterJCLaw, line 10: What is `"e1839a8"` here? It looks suspiciously like a git hash and also very much like the line from the docs about using `-e .`. Surely it's not a magic constant baked into pipenv?

Pipfile
PeterJCLaw: I'm not sure I follow. If this file is (only) used for setting up the (development) virtual environment, then surely that opens the door to the dev environment getting out of step with the deployment environment.

PeterJCLaw:
> you can always use `pipenv shell` which is equivalent to `source venv/bin/activate`.

Ah, that certainly helps somewhat. Could we either clarify that in the README, or (probably better) link to the `pipenv` docs and encourage those who haven't used it before to look for things which fit their existing workflows? (I think `pipenv` is probably still more work for me to use than virtual-envs given that I use `virtualenvwrapper`, though understanding the tool better may clarify the benefits)

> Pipenv is the [standard pattern recommended by the PSF](https://packaging.python.org/new-tutorials/installing-and-using-packages/#installing-pipenv)

Ah, an [appeal to authority](https://yourlogicalfallacyis.com/appeal-to-authority).

> so I don't think you could say it's forcing users to use custom file formats.

That it's the latest recommended pattern doesn't make the file formats any less new. The long standing and familiar (de-facto?) standard is `requirements.txt` (and to an extent `setup.py`). There's lots of independent tooling around those files, while `Pipfile` (and `Pipfile.lock`) files are essentially specific to this new tool (I assume, though if there are other tools which operate on them, please do say so!) .

> Users can still `pip install -e .` the package if they want to.

Could we clarify this? I was definitely under the impression that everyone would *need* to move to `pipenv`.

It's no longer obvious how to run the linting or tests in that scenario. (While I applaud moving the dependencies for these into the `setup.py`, the removal of the related files is arguably breaking. I think this could be mitigated by keeping `script/linting/lint` even if all it does is run the linter under pytest)

PR #9, opened 2017-11-25
Add a util which was very useful today

PeterJCLaw:
This emerged out of testing a bunch of cameras on a Pi and wanting to be sure that:
- they worked with our image capture behaviour, and
- they could successfully capture multiple images with needing some form of manual reset shenanigans

@prophile I've put this in `utils/` and tidied it up a bit, though I'm not sure if that's the best place for it. I see there's already the start of a CLI for the calibration stuff, though I'm not sure what you had in mind for a CLI of the main library.
I think it would be useful to have a CLI which supports the following, though I think most of it would be beyond the scope of this PR:
- plain image capture,
- image capture with April Tags detection (maybe with a visual overlay), and
- just April Tags detection from a saved image

### DIFF
--- a/sb_vision/camera.py
+++ b/sb_vision/camera.py
@@ -12,11 +12,13 @@ from sb_vision.cvcapture import CaptureDevice
 
 from .camera_base import CameraBase
 
+from typing import Tuple
+
 
 class Camera(CameraBase):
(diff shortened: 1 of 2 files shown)


RealOrangeOne:
I kinda think this would be better placed in the `script` directory?

PR #11, opened 2017-12-02
Technet camera warmup

PeterJCLaw:
The interesting commit here is 78ba6121b33d4c16e8dc4094b752392037eb5a27. The preceding commits are already included in #10.

### DIFF
--- a/sb_vision/native/cvcapture.cpp
+++ b/sb_vision/native/cvcapture.cpp
@@ -19,12 +19,22 @@ extern "C" {
 
 #include "opencv2/opencv.hpp"
 
+void warmup(cv::VideoCapture* context) {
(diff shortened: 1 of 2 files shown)


RealOrangeOne (approved):

PeterJCLaw:
I've been thinking about this some more and I think we probably want a way to trigger whether the "warmup" happens from Python. I'm not sure how we'd name that for the capture call though. I'm currently thinking of having either `bool needs_warmup` or maybe even `int frames_to_skip` on both the `cvopen` and `cvcapture` functions.

This was actually what I started with, before falling back to the current change. However as I think we do want to be able to support more than one type of camera I think we need to let Python configure this.

PR #12, opened 2017-12-03
Simplify Vision class

PeterJCLaw:
Separate wrapping of the Apriltag library out from the `Vision` class so that the library lifetimes are all in one place and so the actual intent of the processing is simpler to read.

This also greatly simplifies the initialisation logic of the `Vision` class by moving to lazy properties.


sb_vision/native/apriltag/__init__.py
```diff
@@ -1 +1,98 @@
...
+        )
+
+    def __del__(self):
+        """Deinitialise the library."""
+
+        if self._detector:
+            lib.apriltag_detector_destroy(self._detector)
+
+        if self._working_image:
+            lib.image_u8_destroy(self._working_image)
```
prophile: I'm not wild about `__del__` for resource management, because Python makes no guarantees about when (or even whether) `__del__` is called.

Would be up for a context manager/explicit `close` to do this.

sb_vision/native/apriltag/__init__.py
```diff
@@ -1 +1,98 @@
...
+                and discarded before continuing iteration
+        """
+
+        assert self.image_size == img.size, (
+            "Cannot process images of an incompatible size. Library is "
+            "configured for {}, given image at {}".format(
+                self.image_size,
+                img.size,
+            )
+        )
```
prophile: Would be more in favour of a `ValueError` in this case rather than using `assert`.

`assert`s can be compiled out when Python is run with `-O`, and this isn't an _internal_ assertion about consistent state: from a user's point of view passing an image of the incorrect size is something that _can_ happen.

prophile:
Would you kindly fix the linting errors and we can merge this :)

sb_vision/native/apriltag/__init__.py
PeterJCLaw: c97a296

sb_vision/native/apriltag/__init__.py
PeterJCLaw: 0571a47

PeterJCLaw:
Hnng, apparently I can't leave the TODO note in place, yet I'd very much intended to do so -- it's a longer term thing rather than something for this PR. Would it be better to convert it to an issue?

PeterJCLaw:
Raised TODO as #17.

Adimote:
In a perfect world, we'd allow TODOs but only if they were of the form `//TODO(#17)` where `#17` is the ID of an issue. Have created a trello card for this: https://trello.com/c/TJUQIGNO

PeterJCLaw:
@prophile linting issues fixed; could you take another look?

prophile (approved):

PeterJCLaw:
I can't remember whether I've full-stack tested this; as it's been a while since I created this PR I'll do that before merging.

PR #13, opened 2017-12-03
Move debug code to a module command and update

PeterJCLaw:
This creates the start of a cli structure for sb_vision and moves the vision debug logic into it. That logic is also updated to match its arguments and extended with a selection of useful options.

I'm planning to move the helper from #9 into this structure if we like where this is headed.


sb_vision/cli/debug.py
```diff
@@ -0,0 +1,115 @@
+"""Debug code, load the first video device seen and capture an image."""
+
+import pathlib
+import contextlib
+
+from ..camera import Camera, FileCamera
```
RealOrangeOne, line 6: I have a strong dislike of this parent relative import syntax. Single `.` is fine, but can we change these to absolute

PeterJCLaw: Hrm; I have a preference in favour of relative imports (I recall reading somewhere that they were preferred in packages, though I can't concretely remember why [summoning @prophile in case he knows of a reason]). What's the reasoning for disliking them?

RealOrangeOne: I have no objection to relative import. But only in the same / child modules. Moving up a directory feels odd in Python, likely because the syntax just looks odd IMO (although I get why it's like that)

Adimote: IMO if it's still easily readable and removes the dependency on the name of the package, its an improvement. :shipit:

Adimote (approved):

PR #15, opened 2017-12-05
Update opencv dependencies for stretch

PeterJCLaw:
The package names have changed.

Do we want to spell this such that we support both Jessie and Stretch?


Adimote:
Alistair has already made the same changes, will close this.

PR #16, opened 2017-12-09
Add comment including videoio library

Adimote:
I keep forgetting what library I need to include on my laptop to get my opencv bindings to work, as I'm running a later version where video IO tasks have been split into a separate library. 

Add a comment which indicates what you must do to get it to work.

### DIFF
--- a/sb_vision/native/cvcapture_build.py
+++ b/sb_vision/native/cvcapture_build.py
@@ -30,6 +30,9 @@ ffibuilder.set_source(
         'opencv_core',
         'opencv_highgui',
         'opencv_imgproc',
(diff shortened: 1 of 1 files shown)


kierdavis:
I wonder if this can be made into an automatic test (e.g. by checking the version of opencv with pkg-config)

RealOrangeOne:
We should check if this causes issues on older versions of `opencv`. Perhaps we can just install it by default, and installs which need is will just use it, and those who don't, don't.

PeterJCLaw:
I think the issue is that `videoio` doesn't exist as a separate library in earlier versions.

PR #19, opened 2017-12-14
Add calibration data for the Tecknet camera

Adimote:
Both sizes of marker, support for markers coming soon!


Adimote:
Oops looks like I added the game specific code here, will fix

Adimote:
merging this because it shouldn't break anything and the deadline is tomorrow

PR #20, opened 2017-12-15
Fix invalid comparisons and warnings in cvcapture

Adimote:
The sign of the width and height of an image was being ignored.
Fix that by static_casting all variables.


Adimote:
I forgot I made this PR, would be nice to have, as it stops the `Changing resolution from 1280x720 to  1110302x1080` message (or something to that effect)

PR #21, opened 2018-01-08
Type checking

PeterJCLaw:
Adds (fairly relaxed) type checking.


sb_vision/native/apriltag/detector.py
```diff
@@ -1,14 +1,19 @@
 """AprilTag detector wrapper."""
 
+from typing import Any, Iterator, Tuple
+
 from PIL import Image
 
 from ._apriltag import ffi, lib
 
+if False:
```
RealOrangeOne: er....

setup.cfg
```diff
@@ -29,3 +29,15 @@ multi_line_output = 3
...
 default_section = THIRDPARTY
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+
+
+[mypy]
+warn_unused_ignores = True
+strict_optional = True
+
+disallow_any_generics = True
+# disallow_untyped_calls = True
```
RealOrangeOne: Can we remove this commented code?

PeterJCLaw: 980de983e

sb_vision/native/apriltag/detector.py
PeterJCLaw: This is the way to specify typing-only imports.

PeterJCLaw: Python 3.6 has a better solution (a constant `TYPING` which type-checkers can treat as `True` but which the runtime treats as `False`), but we need Python 3.5 support.

PeterJCLaw: It turns out that the constant I was thinking of ([`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING)) was actually introduced in Python 3.5.2, so we can _probably_ use it. I'll check which patch version of Python is on the Pis before switching over though.

PeterJCLaw: It turns out that the Pis are running python 3.5.3, so this should be fine to use.

PeterJCLaw: 159b74c

PeterJCLaw:
@prophile @RealOrangeOne I think this is ready to go; the only thing I'm not sure about is whether we want to use `flake8-mypy` or not. My current thinking is that we should keep them separate so the lints are fast, but we should probably make a decision once and follow the same pattern for all our repos.

RealOrangeOne:
I'm certainly for _not_ using `flake8-mypy`. Most editors should support type checking ontop of linting, and it can always be installed locally if not. In tests it makes little difference, so I'm all for not using it.

sb_vision/native/apriltag/detector.py
```diff
@@ -1,14 +1,19 @@
 """AprilTag detector wrapper."""
 
+from typing import TYPE_CHECKING, Any, Iterator, Tuple
+
 from PIL import Image
 
 from ._apriltag import ffi, lib
 
+if TYPE_CHECKING:
```
RealOrangeOne, line 9: It looks like there are a few places we're importing for only the sake of type definitions. Is this conditionally really necessary, or is it something overlooked in other locations

PeterJCLaw: Does 8cf929567b86f73b334eb4e0900d80f4871196b5 help?

PeterJCLaw: For clarity: in general I think we should do plain imports of types which are needed for type checking, however the constant is useful when either that would result in the import being circular or the types being imported aren't actually importable at runtime (as in this case).

PeterJCLaw:
I've separated `mypy` from `flake8` in 223dc07, using the same script naming pattern as is established in `runusb`.

PR #22, opened 2018-01-22
Recalibrate images

RealOrangeOne:
Correct the invalid training images for tecknet cameras

## Do not merge
Whilst the images have been updated, the models haven't been trained from this. Mostly because there's no documentation on how to do this. (cc @prophile @Adimote )


PeterJCLaw:
@RealOrangeOne does this just fix the names on the images or actually rewrite their content as well? (I ask as the former will be trivial operations in git, while the latter results in wholly new files)

PeterJCLaw:
@RealOrangeOne I'm pretty sure the remaining work boils down to:
- run `python -m sb_vision.calibration calibrations/tecknet_${DIST}cm -o sb_vision/tecknet_trained_on_${DIST}cm.pkl.xz` for `DIST=10` and `DIST=20`
- change the config somewhere to point at one of those config files

However, @prophile has noted (repeatedly) that we shouldn't need two separate calibrations based on the marker size, so we probably want to drop one of the calibrations (or maybe calibrate with one and test with the other).

RealOrangeOne:
@PeterJCLaw this PR generates whole new files, as that was easier. It shouldnt be too much work to just change the filenames instead of that's preferred (I'm starting to think it might be now).

RE building the images, I fail to see why we _don't_ need 2 models. A 10cm marker will look smaller than a 25cm marker, thus will probably be detected as further away, even if they are infact at the same distance? Unless i'm missing something. I'll wait for a discussion to happen on this with @prophile before generating the new model

PeterJCLaw:
@RealOrangeOne We _could_ use two separate models, though the point is that we don't _need_ to do so.

The logic with two models is roughly:
1. determine id
2. load model for id class
3. use model to determine distance in meters

However that means that the vision model needs to know what size each marker is, meaning that it needs to change if we alter the game.

The logic with one model is roughly:
1. load model
2. use model to determine distance in marker-size units
3. convert from marker-size units to meters

In the latter scenario the last conversion can happen anywhere in the stack, meaning that we can keep the game-specific logic purely in the robot-api if we wanted to.

PeterJCLaw:
With regards to regenerating the images -- it's only worth it if we purge one of the sets of images (otherwise everyone has to download all of them anyway).

PeterJCLaw:
I've just had another look at these images; unfortunately I believe that they're still wrong!

PeterJCLaw:
@RealOrangeOne can we close this?

RealOrangeOne:
I believe so, this looks to either be superseded by #29, or become unnecessary with the future intended opencv integrations

PR #23, opened 2018-01-31
Implement valid coordinate conversion

PeterJCLaw:
Replace the coordinate conversion helper with one which conforms to the usual meanings of polar coordinates and introduce typed named tuples for the coordinate systems so that we get useful representations.

This also adds tests: both unit tests for the conversion util and integration tests that we get the right values for some example images.

### DIFF
--- a/sb_vision/tokens.py
+++ b/sb_vision/tokens.py
@@ -11,7 +11,7 @@ import numpy as np
 from .coordinates import (
     Cartesian,
     cartesian_to_legacy_polar,
-    cartesian_to_spherical
+    cartesian_to_spherical,
 )
 
 if TYPE_CHECKING:
--- a/sb_vision/tokens.py
+++ b/sb_vision/tokens.py
@@ -8,7 +8,11 @@ from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple
 
 import numpy as np
 
-from .coordinates import Cartesian, cartesian_to_polar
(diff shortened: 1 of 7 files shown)


RealOrangeOne (approved):

PeterJCLaw:
@RealOrangeOne I've rearranged this so that we don't break the existing API; could you have another look?

sb_vision/tokens.py
```diff
@@ -222,7 +219,10 @@ def infer_location_from_homography_matrix(
         )
 
         # Polar co-ordinates in the 3D world, relative to the camera
-        self.polar = self.cartesian_to_polar(self.cartesian)
+        self.polar = cartesian_to_legacy_polar(self.cartesian)
```
RealOrangeOne, line 222: what's the point of both `polar` and `legacy_polar`? Surely this is unnecessary?

PeterJCLaw: I want to remove `polar` soon so that internally we only have `legacy_polar`. By having both for a bit we get to so that stepwise, rather than needing to update everything all in one go.

RealOrangeOne (approved):
I can't speak for the maths in `cartesian_to_spherical`, but the code looks ok

PeterJCLaw:
I'm pretty confident of the maths -- I worked out what it should be on pen & paper and then separately worked out the examples which are used in the tests (also by hand).

PR #25, opened 2018-02-03
Marker sizes

Adimote:
make sb-vision marker size aware.

This involves re-adding game_specific.py, but that's okay because we can remove it from robotd now.


sb_vision/game_specific.py
```diff
@@ -0,0 +1,52 @@
+# ******************************************************************************
+#   NOTICE: IF YOU CHANGE THIS FILE PLEASE CHANGE ITS COUNTERPART IN SB_VISION
```
PeterJCLaw: This _is_ "SB_VISION" presumably this should say "robot-api"?

sb_vision/tokens.py
```diff
@@ -183,10 +198,12 @@ def from_apriltag_detection(
         arr = [apriltag_detection.H.data[x] for x in range(9)]
         homography = np.reshape(arr, (3, 3))
 
+        id = apriltag_detection.id
```
PeterJCLaw: We should avoid shadowing a builtin if possible: `id` -> `id_`.

sb_vision/tokens.py
```diff
@@ -130,23 +132,36 @@ def _apply_distance_model_component_to_homography_matrix(
     )
 
 
+# TODO: Don't hard-code this
```
PeterJCLaw: This needs resolving before we can merge.

I suggesting adding a field to the calibrations which records the size.

PeterJCLaw (dismissed):

sb_vision/tokens.py
RealOrangeOne: I think i'd prefer `marker_id`, as it gives a better idea of what's its an id of

sb_vision/tokens.py
```diff
@@ -216,6 +234,7 @@ def infer_location_from_homography_matrix(
             homography_matrix,
             image_size,
             distance_model,
+            marker_size
```
prophile: trailing comma pls

Adimote:
Please re-review

calibrations/tecknet_10cm/files.yaml
```diff
@@ -67,3 +67,4 @@ files:
 - {image: 5.0z1.0x.jpg, x: 1.0, z: 5.0}
 - {image: 5.0z1.5x.jpg, x: 1.5, z: 5.0}
 version: 1
+marker_size: 0.10
```
PeterJCLaw: Trailing newline please.

sb_vision/tokens.py
```diff
@@ -93,6 +94,9 @@ def _get_distance_model(name: str, image_size: Tuple[int, int]) -> Dict[str, Any
             ),
         )
 
+    if 'marker_size' not in calibration:
+        raise ValueError("Calibrations must have a marker_size field")
```
PeterJCLaw: If we're going to require this at the usage of the distance model then we should also require it at the generation of the model -- i.e: in the calibration logic.

PeterJCLaw: Obviously if we change the requirements of the calibration annotations file we'll need to increment the version number.

PeterJCLaw (dismissed):
Please can we have some tests for this.
If we're going to require marker size details from the models when we create them then we should also always output that when calibrating a model.

How were the updated models created -- were they just a modification (i.e: load, edit, save) of the current models or were they re-trained? If re-trained, are we certain that they are equivalent? (While training _should_ be deterministic, I'm overly cautious as I don't get an identical c270 model if I re-train it on master)

sb_vision/calibration/cli.py
```diff
@@ -64,8 +64,7 @@ def main(args=None):
 
     calibration = fit(training_examples)
 
-    if 'marker_size' in config_data:
-        calibration['marker_size'] = config_data['marker_size']
+    calibration['marker_size'] = config_data['marker_size']
```
PeterJCLaw: This will error for config file version 1. We should either alert about that when we check the version number (i.e: not support version 1) or allow it here if (though only if) we're reading a version 2 config.

PeterJCLaw (dismissed):
I'd still like to see some tests.
There's an issue related to reading version 1 config files.

sb_vision/calibration/cli.py
Adimote: fixed in ad62c79

calibrations/example/files.yaml
```diff
@@ -3,4 +3,3 @@ files:
  - image: test.jpg
```
PeterJCLaw, line 3: We should also update the example calibration to version 2.

sb_vision/tokens.py
```diff
@@ -196,7 +215,8 @@ def infer_location_from_homography_matrix(
         *,
         homography_matrix,
         distance_model,
-        image_size
+        image_size,
+        marker_size,
```
PeterJCLaw: Trailing comma here is invalid in Python 3.5 (raises `SyntaxError`). I'm not sure why this doesn't show up on the CI (fixed in 3f4cd37435745a5bd985dedc7235cd8c014d0966).

sb_vision/tokens.py
```diff
@@ -183,10 +200,12 @@ def from_apriltag_detection(
         arr = [apriltag_detection.H.data[x] for x in range(9)]
         homography = np.reshape(arr, (3, 3))
 
+        marker_id = apriltag_detection.id
         instance.infer_location_from_homography_matrix(
             homography_matrix=homography,
             distance_model=distance_model,
             image_size=image_size,
+            marker_size=MARKER_SIZES[marker_id] if marker_id in MARKER_SIZES else None,
```
PeterJCLaw: This could be specified much more simply as `MARKER_SIZES.get(marker_id)` (fixed in 4a36718fc638aa46873ef2592a0caaffb30fe1cd)

sb_vision/tokens.py
```diff
@@ -134,19 +138,32 @@ def _get_cartesian(
     homography_matrix,
     image_size: Tuple[int, int],
     distance_model: str,
+    marker_size: Tuple[float, float]=None,
```
PeterJCLaw: I'm not sure we want this argument to be optional, though we presumably need to allow it to be `None` in case there is no size. 78dd5a6aba5a846ba85b86a82461cc31ad1db2f4 changes it to explicit `Optional[...]` while requiring the argument to be present.

tests/test_marker_sizes.py
```diff
@@ -0,0 +1,46 @@
...
+homography_from_camera = np.array([
+    [7.08277177e-01, -6.30171882e-03, 9.36194251e-01],
+    [1.08272961e-02, 7.11294182e-01, 2.64415168e+00],
+    [2.14327973e-05, 7.37441645e-06, 7.25640594e-03],
+])
+distance_model = 'c270'
+image_resolution = (1280, 720)
+
+
+def test_marker_size():
```
PeterJCLaw: Thanks for adding this, though this test is rather complex to read and doesn't really validate the underlying approach being taken -- it's more just duplicating the logic somewhere else!

I've added some more holistic tests which also validate the underlying approach in 1274f6a8be8210bcf1e5b1f2ae5db4e2a4a8cf96.

PeterJCLaw (dismissed):
There's a trailing comma which is invalid in Python 3.5. It looks like our CI is running Python 3.6 and thus didn't pick up on that.

I've put the fix to that and some other issues in #28.

PeterJCLaw (dismissed):
:shipit:

PeterJCLaw:
Actually, I'll revise that -- I'm still concerned about the regenerated c270 model.

PeterJCLaw (commented):
I think these code changes are ok, I'm concerned that we've regenerated the c270 model, which we're somewhat relying on being non-broken to ensure the vision works. While the regeneration _should_ be deterministic (I _think_), comparing the models shows that there are changes beyond the addition of the key that we're interested in. I'm not sure whether that's something that we really want to do.

Adimote:
@PeterJCLaw I've mitigated this risk by grabbing the old calibration, and manually adding the marker size to the old pkl file

PeterJCLaw (approved):

PR #26, opened 2018-02-03
Pytest invocation from setup.py

kierdavis:
This allows the test suite to be run by executing `python setup.py test`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,3 +48,7 @@ disallow_any_generics = True
 check_untyped_defs = True
 
 ignore_missing_imports = True
+
+
+[aliases]
+test = pytest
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -48,3 +48,7 @@ disallow_any_generics = True
 check_untyped_defs = True
 
 ignore_missing_imports = True
+
+
+[aliases]
```
PeterJCLaw: Could we put this at the top, so as to keep the alphabetical sorting of sections?

PeterJCLaw (approved):
minor suggestion

setup.cfg
kierdavis: 6df539e

RealOrangeOne (approved):

PR #27, opened 2018-02-05
Add calibration comparison command

PeterJCLaw:
This provides a way to compare a trained model against a collection of images which have been annotated with their actual distance values in the same way as the calibration inputs have been.


PeterJCLaw:
Since 90f80f27bb93b70809746f6465d96693d632d214, calibration is no longer a thing.

PR #28, opened 2018-02-08
Fix various issues with marker sizes

PeterJCLaw:
This adds some more tests and fixes some invalid syntax.


RealOrangeOne:
I think this PR may be better done _into_ the original `marker-sizes` branch, so the same code doesnt have to be reviewed twice etc

PeterJCLaw:
@RealOrangeOne oops, I'd intended that.

PeterJCLaw:
This has in fact already been merged into #25.

PR #30, opened 2018-02-13
make pytest truly optional

Adimote:
`pytest` isn't a dependency for the project.

However, if one runs `setup.py install`, it will also install pytest, as it's a requirement for `pytest-runner`, which itself needs to be a setup dependency for `python setup.py test` to call `pytest`.

This PR implements the recommended workaround (see https://pypi.python.org/pypi/pytest-runner#considerations consdierations section) to `setup.py` to make it only be in `setup_requires` when needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,12 @@
+import sys
+
 from setuptools import find_packages, setup
+
+# make pytest-runner be required only when needed
+# This means pytest definitely isn't a dependency if tests aren't run.
+needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
+pytest_runner = ['pytest-runner'] if needs_pytest else []
+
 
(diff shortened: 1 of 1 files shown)


RealOrangeOne (dismissed):
It's be far nicer to use the `test_requires` key to specify things which are only required for tests: https://github.com/sourcebots/robot-api/blob/master/setup.py#L12

RealOrangeOne (approved):

PeterJCLaw:
@Adimote given that building the package also runs the tests, I'm not sure what this gains us. I think we should stick with having the tests run as part of the package build as it prevents us accidentally shipping known-broken things.

PeterJCLaw:
@Adimote do you have an example of a failure which this is trying to resolve?

PR #32, opened 2018-02-17
Remove our thresholding functionality, as apriltags does it itself

ninjafrostpn:


sb_vision/cvcapture.py
```diff
@@ -97,31 +97,3 @@ def close(self) -> None:
...
-        'uint8_t[{}]'.format(width * height),
-    )
-
-    _cvcapture.ffi.memmove(
-        source_buffer,
-        image,
-        len(image),
-    )
-
-    _cvcapture.lib.cvthreshold(
```
PeterJCLaw, line 120: Does this mean we can also remove the native function which provides this?

PeterJCLaw (changes requested):
Conceptually I think this is great, however:
- it would be nice to have some evidence (sample images in various lighting conditions, with detection success rates) that this is a positive change; alternatively can we point at some docs which specify that apriltags does this for us (and that we shouldn't do it ourselves)
- we should ensure the tests pass
- I think there's more code we can remove if we do remove this

sb_vision/cvcapture.py
RealOrangeOne: Reading through, it looks like we can, yes 🎉

Adimote:
@PeterJCLaw Evidence: If you run `python -m sb_vision summarise -m c270 calibrations/c270/*.jpg` on master vs this branch:
 22/55 images are not recognised, whilst in this branch 0/55 are not recognised.

Adimote:
It looks like the tests here are failing because the new thresholding changes where the edges are detected, and the machine learning we used can vary distances greatly with minimal changes to the corner co-ordinates. I'd suggest we wait until #33 is merged, otherwise it'll be wasting effort changing the numbers in this PR just to have them changed again when #33 is merged.

Adimote:
One thing to note is on my laptop it seems to take double the time to process each image if we remove our thresholding...?

PeterJCLaw:
@Adimote:
> Evidence: If you run python -m sb_vision summarise -m c270 calibrations/c270/*.jpg on master vs this branch:
> 22/55 images are not recognised, whilst in this branch 0/55 are not recognised.

Indeed, I was aware of this. However I was hoping for some comparison which are otherwise of the same scene so that things like distance, focus & distractions can be eliminated.

I'd also be interested in seeing the results for a variety of focus settings (also for an otherwise fixed scene; including fixed lighting), and possibly also for a scene which includes two markers with different lighting and different distances (thus different focal effects).

I'm not sure we need _all_ of the above to merge this, though I do think we should do more than just point back at the initial evidence which prompted this investigation (that feels rather circular).

-----
> One thing to note is on my laptop it seems to take double the time to process each image if we remove our thresholding...?

This is concerning, and a strong reason to gather more data (including performance comparison on a Pi) before changing IMO.

Adimote:
I've done some timing of the two separate thresholding methods on a Pi 3. the times are for the entire duration, from starting the capture to getting a list of markers.

Pi thresholing: 0.75-ish
Our thresholding: 0.56-ish

So the added delay to a capture is 210ms, or 35ish%

PeterJCLaw:
@Adimote what do you mean by "Pi" thresholding? Should that read "AprilTags thresholding"?

PeterJCLaw:
@ninjafrostpn meta: git users tend to use dashes (`-`) as separators in their branch names. I've occasionally seen underscores (`_`), though I think I'd generally avoid brackets (they can be hard to type in a shell as they need escaping).

Adimote:
@ninjafrostpn are you thinking of doing this anytime soon? It would be good to get it released today or tomorrow.

sb_vision/cvcapture.py
ninjafrostpn: Sorry, I wasn't quite sure what this meant. Think I've done it now.

ninjafrostpn:
I'm not really sure what to do about it failing the x-value checks; is this just something that will persist until this is merged in?

PeterJCLaw:
Regarding the test failures, I suspect they're a result of the calibrations having been done against images which were pre-thresholded, so I'm not surprised that they've changed.

@ninjafrostpn could you merge in the current `master` branch to this one (now that it contains the new vision mechanism) and see what failures you get then?

sb_vision/cli/debug.py
```diff
@@ -119,7 +109,7 @@ def add_arguments(parser):
         '--save-image',
         dest='output_file',
         type=pathlib.Path,
-        help="Save the processed image (useful with --after-thresholding and "
+        help="Save the processed image (useful with "
```
PeterJCLaw: Please reflow this help text so we don't have the trailing gap here.

ninjafrostpn:
Apologies for thinking that this was simpler than it was. Is this good to go, or does it need more testing performed on it?

Adimote:
I did a quick and dirty blur test by defocussing the webcam various amounts then taking images, I tried hard and couldn't get an image which one algorithm matched and one couldnt. (It might be just that the cameras are very fiddly to accurately move), either way I think they perform similarly in terms of blur

Adimote (approved):

PR #29, opened 2018-02-10
Fixup tecknet reference imagery

PeterJCLaw:
This renames each of the tecknet sample images in an attempt to match the labels to their actual images, mostly without adjusting the set of available labels (I think there's one removal and one addition). It then bundles models to match, though unfortunately the models still aren't particularly great.

This needs review from someone who was present when the images were captured as the label matching is based on assumptions about what distances are actually present in the images that might not be valid.

PR #33, opened 2018-02-18
Implement new vision system

Adimote:
This PR is based off of #29 as it uses some images from the fixed data set in testing. The first commit you should look at is 5bebf1f .

- Added a new vision system using the unofficial python3 opencv bindings.
- Removed the old ML-based vision system (good riddance)
- Use `distance_model` as the new parameter for the camera model xml file generated by opencv
- As the old calibration files are removed too, the default marker size (the size of a marker if its id isn't defined for the game) is now defined in `game_specific.py`
- Update all distance-verifying tests to accept a certain error amount (within 15% of the total distance)


Adimote:
Just trying to get tests to pass before this is ready for review.

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
+        pixel_coords = np.array([list(l) for l in apriltag_detection.p])
+
+        # centre of marker: average the corners
+        pixel_centre = tuple(
+            np.average(np.array([[1, 2], [1, 2], [1, 2]]), axis=0))
+
+        marker_id = apriltag_detection.id
+
         # *************************************************************************
         # NOTE: IF YOU CHANGE THIS, THEN CHANGE ROBOT-API camera.py
```
RealOrangeOne, line 51: You have changed this file, however `robot-api` hasn't been changed (ideally by design). Is this note necessary anymore?

sb_vision/vision.py
```diff
@@ -76,11 +76,12 @@ def process_image(self, img: Image.Image) -> List[Token]:
         """
         img = self.threshold_image(img)
 
-        distance_model = self.camera.distance_model
+        camera_model = self.camera.camera_model
+        detections = list(self.apriltag_detector.detect_tags(img))
```
RealOrangeOne: Why are you casting this to a list, to then just iterate over it? I think keeping it as-is inside the list comprehension is much nicer

tests/test_coordinates_from_file.py
```diff
@@ -5,66 +5,67 @@
...
-            polar_x=1.6704649792860642,
-            polar_y=1.5707963267948966,
-            dist=1.004987562112089,
-        ),
         Spherical(
             rot_x=0,
-            rot_y=math.atan(-0.1),  # or -5.710593137499643 degrees
-            dist=(0.01 + 1)**0.5,  # or 1.004987562112089
+            rot_y=math.atan(-0.1),
+            dist=(0.1 + 1) ** 0.5,
```
PeterJCLaw: If the Cartesian position hasn't changed, then this too should be unchanged.

tests/test_coordinates_from_file.py
```diff
@@ -5,66 +5,67 @@
...
 TEST_IMAGES = [
     (
-        '100cmL10cm.jpg',
+        '1.0z-0.1x.jpg',
         Cartesian(x=-0.1, y=0, z=1),
-        LegacyPolar(
-            polar_x=1.6704649792860642,
-            polar_y=1.5707963267948966,
-            dist=1.004987562112089,
-        ),
```
PeterJCLaw: Why are we removing this? While it's deprecated, it's still supported. We therefore still need to actually have it so should be validating that it outputs the same values.

tests/test_coordinates_from_file.py
```diff
@@ -5,66 +5,67 @@
...
+from sb_vision import Cartesian, FileCamera, Spherical, Vision
 
-CALIBRATIONS = Path(__file__).parent.parent / 'calibrations' / 'c270'
+# Ensure all distance values are within this tolerance
+# (as a percentage of the total distance)
+DIST_PERCENT_TOLERANCE = 0.15
+# Rotation tolerance is independent of distance
+ROTATION_TOLERANCE = 0.075
+
+CALIBRATIONS = Path(__file__).parent.parent / 'calibrations' / 'tecknet_25cm'
```
PeterJCLaw: IMO it would be _much_ better if we could have this still use the c270 samples (I realise that means also calibrating a c270) since we have confidence that the c270 model was correct. That would therefore make it much easier to have confidence that the new calibration mechanism is outputting consistent data.

tests/test_coordinates_from_file.py
PeterJCLaw: Why are you removing the comments here? They were present intentionally to help:
- humans reason about whether values are correct
- make it easier to understand the test failures (since the 'expected' values will be the calculated commented value, rather than the expression)

sb_vision/vision.py
Adimote: Good catch, this was because I iterated over it twice at one point when experimenting, fixed in 523971493427503dd65aa6be8f967d4d25513b0d

tests/test_coordinates_from_file.py
```diff
@@ -5,66 +5,67 @@
...
-            dist=(0.81 + 9)**0.5,  # or 3.132091952673165
+            rot_y=math.atan(1 / 3),
+            dist=(1 + 9) ** 0.5,
         ),
     ),
 ]
 
 
+def assertWithinTolerance(val, expected, tolerance, error_msg=""):
+    assert (expected - tolerance) < val < (expected + tolerance), error_msg
```
PeterJCLaw: I've not used it, but you might enjoy [`pytest.approx`](https://docs.pytest.org/en/latest/builtin.html#comparing-floating-point-numbers)

tests/test_marker_sizes.py
```diff
@@ -9,17 +9,22 @@
...
 
 def assertMarkerDistance(
-    camera: CameraBase,
-    *,
-    marker_sizes: Dict[int, Tuple[float, float]],
-    expected_distance: float
+        camera: CameraBase,
+        *,
+        marker_sizes: Dict[int, Tuple[float, float]],
+        expected_distance: float
```
PeterJCLaw: Why has the indenting here changed?

tests/test_marker_sizes.py
```diff
@@ -9,17 +9,22 @@
...
-EXPECTED_LARGE_DISTANCE = 1.8
-EXPECTED_SMALL_DISTANCE = 0.7
+# The expected error for the distances of markers should be within tolerances
+EXPECTED_LARGE_DISTANCE = 2.5
+EXPECTED_SMALL_DISTANCE = 1
+EXPECTED_TOLERANCE = 0.15
+
+
+def assertWithinToleranceRatio(val, expected, tolerance, error_msg=""):
+    assert expected * (1 - tolerance) < val < expected * (1 + tolerance), \
```
PeterJCLaw: I've not used it, but you might enjoy [`pytest.approx`](https://docs.pytest.org/en/latest/builtin.html#comparing-floating-point-numbers)

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
 
     @classmethod
     def from_apriltag_detection(
-        cls,
-        apriltag_detection: 'ApriltagDetection',
-        image_size: Tuple[int, int],
-        distance_model: Optional[str],
+            cls,
+            apriltag_detection: 'ApriltagDetection',
+            camera_model: Optional[str],
```
PeterJCLaw: Please can we only indent these four spaces? (i.e: unchanged from how they were indented)

sb_vision/tokens.py
```diff
@@ -1,176 +1,34 @@
...
-
 class Token:
     """Representation of the detection of one token."""
 
-    def __init__(self, id: int, certainty: float=0.0) -> None:
+    def __init__(self,
+                 id: int,
+                 certainty=0,
+                 pixel_coords: Iterable[Tuple[float, float]] = (),
+                 pixel_centre: Tuple[float, float] = (0, 0)) -> None:
```
PeterJCLaw: Please can we make the wrapping here consistent with the rest of the repo's multi-line signatures? (see `from_apriltag_detection` for an example)

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
+            apriltag_detection: 'ApriltagDetection',
+            camera_model: Optional[str],
     ) -> 'Token':
         """Construct a Token from an April Tag detection."""
+
+        pixel_coords = np.array([list(l) for l in apriltag_detection.p])
+
+        # centre of marker: average the corners
+        pixel_centre = tuple(
+            np.average(np.array([[1, 2], [1, 2], [1, 2]]), axis=0))
```
PeterJCLaw: Please can we make the wrapping here consistent with the rest of the repo's multi-line signatures? Essentially the pattern is that (if wrapping at all) an expression is either wrapped or not, and that its preceding and following items go on separate lines. The aim is to minimise the diff if another line/argument/expression/etc. needs to be added.

Here that means you _could_ wrap this like this:
``` python
        pixel_centre = tuple(
            np.average(np.array([[1, 2], [1, 2], [1, 2]]), axis=0),
        )
```
though the following would be nicer:
``` python
        pixel_centre = tuple(np.average(
            np.array([[1, 2], [1, 2], [1, 2]]),
            axis=0,
        ))
```
You could also do:
``` python
    pixel_centre = tuple(np.average(np.array([
        [1, 2],
        [1, 2],
        [1, 2]
    ]), axis=0))
```
though that's less clear about where the `axis=0` applies; if you like that list wrapping, you could combine both:
``` python
        pixel_centre = tuple(np.average(
            np.array([
                [1, 2],
                [1, 2],
                [1, 2]
            ]),
            axis=0,
        ))
```

PeterJCLaw: It would be excellent if we could make this a typed `NamedTuple` rather than a plain tuple.

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
-        apriltag_detection: 'ApriltagDetection',
-        image_size: Tuple[int, int],
-        distance_model: Optional[str],
+            cls,
+            apriltag_detection: 'ApriltagDetection',
+            camera_model: Optional[str],
     ) -> 'Token':
         """Construct a Token from an April Tag detection."""
+
+        pixel_coords = np.array([list(l) for l in apriltag_detection.p])
```
PeterJCLaw: Should this also be being converted to a `tuple`? (If so, can we make it a `typing.NamedTuple`?)

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,104 @@
...
+    camera_matrix = calibrations['cameraMatrix']
+    distance_coefficents = calibrations['dist_coeffs']
+    return camera_matrix, distance_coefficents
+
+
+def calculate_transforms(
+        marker_size: Tuple[float, float],
+        pixel_coords: np.ndarray,
+        camera_matrix: np.ndarray,
+        distance_coefficients: np.ndarray):
```
PeterJCLaw: See previous comments about wrapping signatures and indenting.

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,104 @@
...
+    :param marker_size: size of the marker
+    :param pixel_coords: pixel co-ordinates of the corners of the marker
+        (clockwise around the marker)
+    :param camera_matrix: calibration matrix for the camera
+    :param distance_coefficients: distance calibration for the camera
+    :return: translation and orientation of the marker
+    """
+    x, y = marker_size
+    x /= 2
+    y /= 2
```
PeterJCLaw: Why diving by two? (Is there a spelling of either the original variable names or the new ones which would clarify this without a comment?)

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
-            distance_model=distance_model,
-            image_size=image_size,
-            marker_size=MARKER_SIZES.get(marker_id),
-        )
+        # We don't set coordinates in the absence of a
+        # camera model.
+        if camera_model:
+            builtin_models_dir = Path(__file__).parent
+            model_file = builtin_models_dir / '{}_calibration.xml'.format(
+                camera_model)
```
PeterJCLaw: See previous notes about wrapping.

setup.py
```diff
@@ -24,6 +24,8 @@
         'sb_vision/native/apriltag/apriltag_build.py:ffi',
     ],
     install_requires=[
+        'lxml',
+        'opencv-python',
```
PeterJCLaw: Have you checked that these are available on the Pis?

Additionally, how do you intend to deploy them given that the Pis have no internet connection, nor a mechanism to install arbitrary new packages?

sb_vision/tokens.py
```diff
@@ -179,65 +37,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
-        distance_model,
-        image_size,
-        marker_size: Optional[Tuple[float, float]]
+    def infer_location_from_transforms(
+            self,
+            *,
+            orientation: Tuple[float, float, float],
+            translation: Tuple[float, float, float]
     ):
         """Infer coordinate information from a homography matrix."""
```
PeterJCLaw: Presumably this docstring also needs updating?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,105 @@
...
+def _get_text(element: etree.Element):
+    """parse an xml tag with space-separated variables"""
+    text = []
+    for e in element.itertext():
+        if e.strip():
+            text += re.split('\s+', re.sub('\n', '', e.strip()))
+    return text
+
+
+def get_calibration(file_name: Path) -> Dict:
```
PeterJCLaw: This will (should) be failing type checking -- please can we specify the details of this `Dict`. If they're too complex, consider using a `typing.NamedTuple` instead.

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,105 @@
...
+        for element in root:
+            if 'type_id' in element.attrib and element.attrib[
+                 'type_id'] == 'opencv-matrix':
+                rows, cols = int(element.find('rows').text), int(
+                    element.find('cols').text)
+                data_type = element.find('dt').text
+                values = _get_text(element.find('data'))
+                if data_type == 'd':  # doubles
+                    data = np.reshape(np.array([float(v) for v in values]),
+                                      (rows, cols))
```
PeterJCLaw: What's the benefit of going into numpy here? If there's precision that we're aiming for (numpy's `float64` is more precise than `float` is IIRC), then we should be using `float64` to parse the string. Otherwise IMO we should stick to Python types here (not least as it makes the type signatures much easer to reason about).

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,105 @@
...
+from pathlib import Path
+from typing import Tuple, Dict
+
+import cv2
+from lxml import etree
+import re
+import numpy as np
+
+
+def _get_text(element: etree.Element):
```
PeterJCLaw: This doesn't really return "text"; is there a better name for it?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,105 @@
...
+                values = _get_text(element)
+                data = [float(v) for v in values]
+            else:
+                data = _get_text(element)
+            calibrations[element.tag] = data
+    return calibrations
+
+
+@functools.lru_cache()
+def load_camera_calibrations(file_name: Path) -> Tuple[np.ndarray, np.ndarray]:
```
PeterJCLaw: Note that typing of numpy types doesn't actually work -- they're treated as `Any`, so if we can avoid using their array types and instead use built-in types that might be better.

sb_vision/vision.py
```diff
@@ -77,7 +77,7 @@ def process_image(self, img: Image.Image) -> List[Token]:
         img = self.threshold_image(img)
 
         camera_model = self.camera.camera_model
-        detections = list(self.apriltag_detector.detect_tags(img))
+        detections = self.apriltag_detector.detect_tags(img)
```
PeterJCLaw: I think Jake's point was that we could leave this in the comprehension below entirely. This would both minimise the diff and also keep the clarity that the previous spelling had.

PeterJCLaw (commented):
I realise this is work-in progress, so some of the internal variances might be things already on your list of TODOs; hopefully this is useful input anyway.

tests/test_coordinates_from_file.py
Adimote: I don't think we should calibrate a c270 just so we don't need to change these tests. Calibrating has to be done manually using the interactive opencv script, and is thus more effort than it's worth.

tests/test_coordinates_from_file.py
Adimote: Oops that's a miss-calculation, forgot these values are squared. I'll make it more explicit.

tests/test_coordinates_from_file.py
Adimote: fixed in f1e01541de6db6fb570e973fc66f996620b68ef1

tests/test_coordinates_from_file.py
Adimote: re-added them in 5125186fc86040866cfad745234d4c3709caeb81

tests/test_marker_sizes.py
Adimote: fixed in f1e0154

tests/test_marker_sizes.py
Adimote: I have literally no clue how this happened, fixed in 587d6e51a7fe710a4b5bc55d5bfa1d7290e20d8

sb_vision/distance_finding.py
Adimote: fixed in 50d2a455a31dd63df0c29fa0c0a5c1bd22867506

Adimote: and a13fb5d65b1cf48ebb0c763e5bcef5b5fb97f0e3

sb_vision/distance_finding.py
Adimote: fixed in df55c154cb73cbd1c36c94c59349c9ef65c9ba41

sb_vision/distance_finding.py
Adimote: fixed in 49cb34f7b313758ff45c1577fe06faa61edfb562

sb_vision/tokens.py
Adimote: Fixed in 87a0a4e135dbcfd753baf6e1d7ec66b5149b71b1

sb_vision/tokens.py
Adimote: fixed in e2f6792ffc8a6a3dbf01d393607269865613f8d6

sb_vision/tokens.py
Adimote: then re-fixed in e2f6792ffc8a6a3dbf01d393607269865613f8d6

sb_vision/tokens.py
Adimote: Yes, which also removed the need for some typing.cast calls before! (yay!) fixed in dc67a9179aeba26da69161bf8c6f87c172ca342f

sb_vision/tokens.py
Adimote: It happens to be that the changes I made here doesn't affect robot-api's version. If I were to add a new value (i.e. orientation), I'd need to add it in robot-api, so I think we still need have a need for the message

sb_vision/tokens.py
Adimote: fixed in e5b8775e2f4e58c465ffde8e8fe64b7753ed7082

sb_vision/tokens.py
Adimote: fixed in 356c3f4cafeb5620412cfcbbe24e310f875188dd

setup.py
Adimote: we'll need to include their debs in the `pi-image` update. Good point but I don't think it blocks this PR.

tests/test_marker_sizes.py
Adimote: Actually, I do. It seems Pycharm defaults to this when you autoformat with it

tests/test_coordinates_from_file.py
Adimote: re-added in df9df93db918fb402c23ff81e337acbd223d02f6

Adimote:
Okay, this is ready for proper review now

sb_vision/distance_finding.py
Adimote: fixed in 304abe6

sb_vision/tokens.py
PeterJCLaw: I'm not sure that that's strictly true, and certainly nowhere near as dramatic as this comment appears.

_Adding_ a value here _makes it possible_ to also expose that value in robot-api. _Removing_ a value here is (hopefully obviously) a breaking change, so we'd need to modify the consumers of the library anyway.

I'd be in favour of reducing the shoutyness of this comment. However, from what I can see the comment itself doesn't change here, so that would be a separate PR anyway.

sb_vision/tokens.py
Adimote: Thanks for pointing this out, also fun fact: that code is actually a bug in itself, it shouldn't statically return the average of 1 and 2, it should return the average of the corner pixels

tests/test_coordinates_from_file.py
PeterJCLaw: @Adimote Out of curiosity: what are the error messages like from `pytest.approx`? Are they useful in helping understand the failure?

sb_vision/tokens.py
Adimote: The decision to wrap this way is not conscious, it's just a default in pycharm

tests/test_coordinates_from_file.py
```diff
@@ -4,38 +4,45 @@
...
+            polar_x=1.6697,
+            polar_y=1.5740,
+            dist=(0.1**2 + 1**2) ** 0.5,  # or 1.004987562112089
         ),
         Spherical(
             rot_x=0,
-            rot_y=math.atan(-0.1),  # or -5.710593137499643 degrees
-            dist=(0.01 + 1)**0.5,  # or 1.004987562112089
+            rot_y=math.atan(-0.1),  # or -0.09966865249116204 radians
+            dist=(0.1**2 + 1**2) ** 0.5,  # or 1.004987562112089
```
PeterJCLaw: If these lines aren't actually changing, can we just leave them as they were? Adding diff-noise to an already _huge_ PR is massively undesirable.

tests/test_coordinates_from_file.py
```diff
@@ -4,38 +4,45 @@
...
-        '100cmL10cm.jpg',
+        '1.0z-0.1x.jpg',
         Cartesian(x=-0.1, y=0, z=1),
         LegacyPolar(
-            polar_x=1.6704649792860642,
-            polar_y=1.5707963267948966,
-            dist=1.004987562112089,
+            polar_x=1.6697,
+            polar_y=1.5740,
+            dist=(0.1**2 + 1**2) ** 0.5,  # or 1.004987562112089
```
PeterJCLaw: If these lines aren't actually changing, can we just leave them as they were? Adding diff-noise to an already _huge_ PR is massively undesirable.

tests/test_coordinates_from_file.py
```diff
@@ -4,38 +4,45 @@
...
-            dist=3.132091952673165,
+            polar_x=1.1922,
+            polar_y=1.6032,
+            dist=(1**2 + 3**2) ** 0.5,  # or 3.1622776601683795
         ),
         Spherical(
             rot_x=0,
-            rot_y=math.atan(-0.9 / 3),  # or -16.69924423399362 degrees
-            dist=(0.81 + 9)**0.5,  # or 3.132091952673165
+            rot_y=math.atan(1 / 3),  # or 0.3217505543966422 radians
```
PeterJCLaw: Could we also have this in degrees? I don't have any intuition about how big an angle 0.32 radians is.

tests/test_coordinates_from_file.py
```diff
@@ -47,24 +54,27 @@
...
 def test_image_coordinates(photo, expected_cartesian, expected_polar, expected_spherical):
     """Make sure that this particular file gives these particular tokens."""
-    camera = FileCamera(CALIBRATIONS / photo, distance_model='c270')
+    camera = FileCamera(CALIBRATIONS / photo, camera_model='C016')
     vision = Vision(camera)
     token, = vision.snapshot()
 
     x, y, z = token.cartesian
+    rot_x, rot_y, dist = token.spherical
+    polar_x, polar_y, polar_dist = token.legacy_polar
```
PeterJCLaw: Do we need to move these lines up to the top? Leaving them interleaved keeps the context for each triplet of assertions (and reduces diff noise).

tests/test_marker_sizes.py
```diff
@@ -4,15 +4,23 @@
...
-# bit off; note that they still preserve the ratio of 1:2.5 (0.7 : 1.75 then rounded)
-EXPECTED_LARGE_DISTANCE = 1.8
-EXPECTED_SMALL_DISTANCE = 0.7
+# The expected error for the distances of markers should be within tolerances
+EXPECTED_LARGE_DISTANCE = 2.5
+EXPECTED_SMALL_DISTANCE = 1
+EXPECTED_TOLERANCE = 0.15
+
+
+def assertWithinToleranceRatio(val, expected, tolerance, error_msg=""):
```
PeterJCLaw: This is unused.

tests/test_coordinates_from_file.py
```diff
@@ -47,24 +54,27 @@
...
-    assert expected_cartesian.y == round(y, 3), "Wrong y-coordinate"
-    assert expected_cartesian.z == round(z, 3), "Wrong z-coordinate"
-
-    polar_x, polar_y, dist = token.legacy_polar
+    tolerance = dist * DIST_PERCENT_TOLERANCE
 
-    assert round(expected_polar.polar_x, 3) == round(polar_x, 3), "Wrong x angle"
-    assert round(expected_polar.polar_y, 3) == round(polar_y, 3), "Wrong y angle"
-    assert round(expected_polar.dist, 3) == round(dist, 3), "Wrong distance"
+    assert x == approx(expected_cartesian.x, abs=tolerance), "Wrong x-coordinate"
```
PeterJCLaw: I'm not sure if I really like this idea, but it _might_ make the assertions easier if we did something like `approx_dist = functools.partial(approx, abs=dist * DIST_PERCENT_TOLERANCE)`.

That would make the test linses a bit shorter while removing some of the duplication. It would also clarify that the tolerance is calibrated against a _distance_.

tests/test_coordinates_from_file.py
```diff
@@ -47,24 +54,27 @@
...
     x, y, z = token.cartesian
+    rot_x, rot_y, dist = token.spherical
+    polar_x, polar_y, polar_dist = token.legacy_polar
 
-    assert expected_cartesian.x == round(x, 3), "Wrong x-coordinate"
-    assert expected_cartesian.y == round(y, 3), "Wrong y-coordinate"
-    assert expected_cartesian.z == round(z, 3), "Wrong z-coordinate"
-
-    polar_x, polar_y, dist = token.legacy_polar
+    tolerance = dist * DIST_PERCENT_TOLERANCE
```
PeterJCLaw: We should probably define the tolerance in terms of the _expected_ rather than _actual_ value. I realise it ought to make little practical difference, it just seems more "obviously correct" to me.

setup.py
PeterJCLaw: I very much think that this _does_ block this PR. We shouldn't be merging anything which we're not in a position to actually deploy.

If we can't solve that, but still want to get this support into `master`, one option might be to support both the old mechanism and the new one, with some run-time switch between the two (such that the imports happen after the choice). That's a bit nasty, but would allow us to migrate slowly.

I think if we can get debian packages for these dependencies and include them in an update file, then as long as `sb-vision` depends on them they'll get installed.

`lxml` is already [packaged for stretch](https://packages.debian.org/stretch/python3-lxml) so is likely available for the Pis, though I  can't find evidence that `opencv-python` is. If it comes to it, we can probably package it ourselves, though that's not ideal.

The other alternative would be to move any use of these dependencies to the calibration step, though I don't think that that's possible for `opencv-python` which is unfortunate given that that's the package least available.

I really don't want us to merge this and then have to revert it because we can't deploy it.

sb_vision/vision.py
```diff
@@ -76,11 +76,12 @@ def process_image(self, img: Image.Image) -> List[Token]:
...
 
-        distance_model = self.camera.distance_model
+        camera_model = self.camera.camera_model
+        detections = self.apriltag_detector.detect_tags(img)
 
         tokens = [
-            Token.from_apriltag_detection(x, img.size, distance_model)
-            for x in self.apriltag_detector.detect_tags(img)
+            Token.from_apriltag_detection(x, camera_model)
+            for x in detections
```
PeterJCLaw: What's the point of extracting `detections` to a variable here? (I'm guessing this was a debugging change which has leaked into the final PR?)

sb_vision/tokens.py
```diff
@@ -179,65 +40,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
-        self.homography_matrix = homography_matrix
-
-        # We don't set cartesian coordinates in the absence of a
-        # distance model.
-        if distance_model is None:
-            return
-
+        """Infer coordinate information from the given transformations."""
         # Cartesian Co-ordinates in the 3D World, relative to the camera
-        # (as opposed to somehow being compass-aligned)
```
PeterJCLaw: Why remove this comment? (Please don't modify unrelated things just because you're near them, especially in an already large PR)

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+    distance_coefficents = calibrations['dist_coeffs']
+    return camera_matrix, distance_coefficents
+
+
+def calculate_transforms(
+    marker_size: Tuple[float, float],
+    pixel_coords: List[Coordinate],
+    camera_matrix: List[List[float]],
+    distance_coefficients: List[List[float]],
+):
```
PeterJCLaw: This needs a return type. I'm surprised that `mypy` isn't complaining about that.

sb_vision/tokens.py
```diff
@@ -179,65 +40,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
-        self.cartesian = _get_cartesian(
-            homography_matrix,
-            image_size,
-            distance_model,
-            marker_size,
-        )
+        self.cartesian = Cartesian(*translation)
```
PeterJCLaw: Can we push the `Cartesian` typing upwards here? From what I can tell, it's generated in `calculate_transforms` -- is there any reason not to have that return a `Cartesian` rather than a plain tuple?

For strict correctness there's an argument that `calculate_transforms` should return a `Translation` type which has the same definition as `Cartesian` and would be a no-op conversion. One could even extend this to argue that what the `Token`s actually have is a `Translation` rather than a `Cartesian`, so maybe we just accept that that's what `Cartesian` means anyway and just use that (rather than introducing a new type)?

I'm not sure we need to introduce a `Translation` type unless you particularly feel you want to (I certainly think that changing `Cartesian` is probably more effort than it's worth given that it's referenced all over the place), but I would like to see us push the type further up the call-stack.

sb_vision/tokens.py
```diff
@@ -179,65 +40,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
     # noinspection PyAttributeOutsideInit
-    def infer_location_from_homography_matrix(
+    def infer_location_from_transforms(
         self,
         *,
-        homography_matrix,
-        distance_model,
-        image_size,
-        marker_size: Optional[Tuple[float, float]]
+        orientation: Tuple[float, float, float],
```
PeterJCLaw: `orientation` seems unused. Is that intentional? (If so, why's it here?)

sb_vision/tokens.py
```diff
@@ -179,65 +40,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
+
+            instance.infer_location_from_transforms(
+                orientation=orientation,
+                translation=translation,
+            )
         return instance
 
     # noinspection PyAttributeOutsideInit
-    def infer_location_from_homography_matrix(
+    def infer_location_from_transforms(
```
PeterJCLaw: Is there any inference actually happening here now? (I can't see any).

Would it make sense to either push more of the logic from the `if camera_model` (though obviously not the condition) block down into the method, or just scrap the method and everything inline in `from_apriltag_detection`?

Alternatively is there a better name for this method?

sb_vision/tokens.py
```diff
@@ -179,65 +40,68 @@ def __init__(self, id: int, certainty: float=0.0) -> None:
...
+        # We don't set coordinates in the absence of a
+        # camera model.
+        if camera_model:
+            builtin_models_dir = Path(__file__).parent
+            model_file = builtin_models_dir / '{}_calibration.xml'.format(
+                camera_model,
+            )
+            camera_matrix, distance_coefficents = load_camera_calibrations(
+                model_file,
+            )
```
PeterJCLaw: Would it make sense to bundle these lines into a separate util? As it stands, these are quite low-level operations in what is otherwise a function focussing on higher level things.

Maybe something like this:
``` python
@functools.lru_cache()
def _get_camera_calibrations(camera_name: str) -> ...:
    if camera_model is None:
        raise ValueError("Cannot get camera calibration of None")

    builtin_models_dir = Path(__file__).parent
    model_file = builtin_models_dir / '{}_calibration.xml'.format(camera_name)
    return load_camera_calibrations(model_file)
```

sb_vision/tokens.py
```diff
@@ -13,164 +10,28 @@
...
 class Token:
     """Representation of the detection of one token."""
 
-    def __init__(self, id: int, certainty: float=0.0) -> None:
+    def __init__(
+        self,
+        id: int,
+        certainty=0,
+        pixel_coords: Iterable[Tuple[float, float]] = (),
+        pixel_centre: Tuple[float, float] = (0, 0),
```
PeterJCLaw: Surely we have more precise types for the `pixel_coords` and `pixel_centre` parameters?

sb_vision/tokens.py
```diff
@@ -13,164 +10,28 @@
...
-
 class Token:
     """Representation of the detection of one token."""
 
-    def __init__(self, id: int, certainty: float=0.0) -> None:
+    def __init__(
+        self,
+        id: int,
+        certainty=0,
+        pixel_coords: Iterable[Tuple[float, float]] = (),
```
PeterJCLaw: I don't think we should allow this value to be empty or omitted -- previously it was always present and always had four tuples in the list. I realise we can't enforce the number of items through types, though we should require that it be a `List` (otherwise is breaking) and expect that it contains the right number of items.

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # create the rectangle representing the marker in 3D
+    object_points = np.array([
+        [width_from_centre, height_from_centre, 0],
+        [width_from_centre, -height_from_centre, 0],
+        [-width_from_centre, -height_from_centre, 0],
+        [-width_from_centre, height_from_centre, 0],
+    ])
+
+    _, orientation_vector, translation_vector = cv2.solvePnP(
```
PeterJCLaw: Curiosity: what's the extra variable we're ignoring here?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+        object_points,
+        np.array(pixel_coords),
+        np.array([np.array(xi) for xi in camera_matrix]),
+        np.array([np.array(xi) for xi in distance_coefficients]),
+    )
+    translation_vector = tuple(v[0] for v in translation_vector)
+    orientation_vector = tuple(v[0] for v in orientation_vector)
```
PeterJCLaw: Curiosity, though a comment explaining would likely be useful: what's the shape of the `translation_vector` and in particular what are we throwing away from the `v`s when we get `v[0]`?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+        [width_from_centre, -height_from_centre, 0],
+        [-width_from_centre, -height_from_centre, 0],
+        [-width_from_centre, height_from_centre, 0],
+    ])
+
+    _, orientation_vector, translation_vector = cv2.solvePnP(
+        object_points,
+        np.array(pixel_coords),
+        np.array([np.array(xi) for xi in camera_matrix]),
+        np.array([np.array(xi) for xi in distance_coefficients]),
```
PeterJCLaw: Why `xi`? (was there a previous `x` in scope you were avoiding conflicting with?)

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+):
+    """
+    Calculate the position of a marker.
+
+    given the pixel co-ordinates of the corners and the calibrations
+    of the camera.
+
+    :param marker_size: size of the marker
+    :param pixel_coords: pixel co-ordinates of the corners of the marker
+        (clockwise around the marker)
```
PeterJCLaw: clockwise from what starting point?

sb_vision/distance_finding.py
PeterJCLaw: In fact, I don't think we need to do this looping ourselves -- I think we can just pass the lists of lists direct to `np.array`:
``` python
>>> np.array([
>>>     [1, 2],
>>>     [2, 3],
>>> ])
array([[1, 2],
       [2, 3]])
```

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+def calculate_transforms(
+    marker_size: Tuple[float, float],
+    pixel_coords: List[Coordinate],
+    camera_matrix: List[List[float]],
+    distance_coefficients: List[List[float]],
+):
+    """
+    Calculate the position of a marker.
+
+    given the pixel co-ordinates of the corners and the calibrations
```
PeterJCLaw: Can we start this sentence with a capital letter please.

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+import functools
+import re
+from pathlib import Path
+from typing import Any, Dict, List, NamedTuple, Tuple
+
+import cv2
+import numpy as np
+from lxml import etree
+
+Coordinate = NamedTuple('PixelCoordinate', [('x', float), ('y', float)])
```
PeterJCLaw: I'm surprised that something isn't erroring about this having a different declared name than its symbol.

PeterJCLaw: This should probably be declared in "coordinates.py".

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def _get_values_from_xml_element(element: etree.Element) -> List[str]:
+    """Parse an xml tag with space-separated variables."""
+    text = []  # type: List[str]
+    for e in element.itertext():
+        if e.strip():
+            text += re.split('\s+', re.sub('\n', '', e.strip()))
```
PeterJCLaw: Does the inner here really need to be `re.sub`? It looks like a plain [`str.replace`](https://docs.python.org/3/library/stdtypes.html#str.replace) would do just fine.

PeterJCLaw: Consider avoiding duplication of `e.strip()`.

PeterJCLaw: Consider using r-literals for the strings being passed to regex methods to ensure clarity around the backslash escapes:
``` python
>>> print(r'\\')
\\
>>> print('\\')
\
```

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+    :param file_name: name of xml file to parse
+    :return: calibrations from the file in a dictionary
+    """
+    calibrations = {}
+    with file_name.open() as file:
+        tree = etree.parse(file)
+        root = tree.getroot()
+        for element in root:
+            if 'type_id' in element.attrib and element.attrib[
+               'type_id'] == 'opencv-matrix':
```
PeterJCLaw: Not sure I'm keen on wrapping the index over multiple lines like this. Could we have something a bit more in line with how wrapping is done elsewhere in this project?
Alternatively, if `.attrib` is sufficiently dict-like could we spell this as `element.attrib.get('type_id') == 'opencv-matrix'` which I suspect wouldn't need wrapping?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+                rows, cols = int(element.find('rows').text), int(
+                    element.find('cols').text)
```
PeterJCLaw: The wrapping here seems incredibly strained -- maybe just put the `rows` and `cols` on separate lines as separate expressions?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+                data_type = element.find('dt').text
+                values = _get_values_from_xml_element(element.find('data'))
+                if data_type == 'd':  # doubles
+                    data = np.reshape([float(v) for v in values],
+                                      (rows, cols)).tolist()
```
PeterJCLaw: Can we make this wrapping fit the style of the repo?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+        root = tree.getroot()
+        for element in root:
+            if 'type_id' in element.attrib and element.attrib[
+               'type_id'] == 'opencv-matrix':
+                rows, cols = int(element.find('rows').text), int(
+                    element.find('cols').text)
+                data_type = element.find('dt').text
+                values = _get_values_from_xml_element(element.find('data'))
+                if data_type == 'd':  # doubles
+                    data = np.reshape([float(v) for v in values],
```
PeterJCLaw: Maybe pull the `_get_values_from_xml_element` inside this `if` since it's not needed on the `else` path?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+            # Integer tag names
+            elif element.tag in ['framesCount', 'cameraResolution']:
+                values = _get_values_from_xml_element(element)
+                data = [int(v) for v in values]
+            elif element.tag in ['avg_reprojection_error']:
+                values = _get_values_from_xml_element(element)
+                data = [float(v) for v in values]
+            else:
+                data = _get_values_from_xml_element(element)
+            calibrations[element.tag] = data
```
PeterJCLaw: IMO newlines after each sub-block would help clarify the if/elif/else here:
``` python
            elif element.tag in ['avg_reprojection_error']:
                values = _get_values_from_xml_element(element)
                data = [float(v) for v in values]

            else:
                data = _get_values_from_xml_element(element) 

        calibrations[element.tag] = data
```
(etc.)

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+                rows, cols = int(element.find('rows').text), int(
+                    element.find('cols').text)
+                data_type = element.find('dt').text
+                values = _get_values_from_xml_element(element.find('data'))
+                if data_type == 'd':  # doubles
+                    data = np.reshape([float(v) for v in values],
+                                      (rows, cols)).tolist()
+                else:
+                    raise ValueError('Invalid data type in xml')
+            # Integer tag names
```
PeterJCLaw: Is this comment adding anything?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+        tree = etree.parse(file)
+        root = tree.getroot()
+        for element in root:
+            if 'type_id' in element.attrib and element.attrib[
+               'type_id'] == 'opencv-matrix':
+                rows, cols = int(element.find('rows').text), int(
+                    element.find('cols').text)
+                data_type = element.find('dt').text
+                values = _get_values_from_xml_element(element.find('data'))
+                if data_type == 'd':  # doubles
```
PeterJCLaw: Maybe push this check to the top of the block so we either early-raise or do the work?

Combining this with the other suggestions yields:
``` python
            if element.attrib.get('type_id') == 'opencv-matrix':
                data_type = element.find('dt').text
                if data_type != 'd':  # doubles
                    raise ValueError('Invalid data type in xml')

                rows = int(element.find('rows').text)
                cols = int(element.find('cols').text)

                values = _get_values_from_xml_element(element.find('data'))
                data = np.reshape(
                    [float(v) for v in values],
                    (rows, cols),
                ).tolist()

            elif element.tag in ['framesCount', 'cameraResolution']:
                ...
```

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+        if e.strip():
+            text += re.split('\s+', re.sub('\n', '', e.strip()))
+    return text
+
+
+def get_calibration(file_name: Path) -> Dict[str, Any]:
+    """
+    Parse a calibration xml generated by the camera calibration tool.
+
+    (see https://docs.opencv.org/3.4.0/d7/d21/tutorial_interactive_calibration.html)
```
PeterJCLaw: It's great to have this link here, though we'll also need some wider documentation of this. Please could you update the calibration section of the README to match the new procedure? (A brief description then pointing at this link is fine; excellent would be also including brief instructions on validating the resulting calibration and mentioning which updates need to be made to include the calibration in the python/debian packages)

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+def _get_values_from_xml_element(element: etree.Element) -> List[str]:
+    """Parse an xml tag with space-separated variables."""
+    text = []  # type: List[str]
+    for e in element.itertext():
+        if e.strip():
+            text += re.split('\s+', re.sub('\n', '', e.strip()))
+    return text
+
+
+def get_calibration(file_name: Path) -> Dict[str, Any]:
```
PeterJCLaw: This function contains conversions for a bunch of things we're not using. Specifically we need the keys `'cameraMatrix'` `'dist_coeffs'` yet the code also seems to cope with translating a bunch of others.

Are you expecting that we'll need those other things at some point, or could we drop the redundant code?

Note: this may make some of the following review points redundant.

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,121 @@
...
+                values = _get_values_from_xml_element(element)
+                data = [float(v) for v in values]
+            else:
+                data = _get_values_from_xml_element(element)
+            calibrations[element.tag] = data
+    return calibrations
+
+
+@functools.lru_cache()
+def load_camera_calibrations(file_name: Path) -> Tuple[List[List[float]],
```
PeterJCLaw: Thinking again about the types this returns, I wonder whether there's a better way to use the type system to help with the return types here. As it stands we get some benefit, but it's limited.

One thing I'd wondered about doing (for the homography matrix) in the previous code was to introduce custom types for the tricky types we have from external sources (such as numpy). I think these would need explicit casts into them, though I think that that's ok as they're opaque to most of the things which are just passing them around anyway.

Certainly, it could help here (and in `calculate_transforms` below) since we're not just after any list of lists of floats, we're actually after a specific shape of them. As such we could declare our own `CameraMatrix` and `DistanceCoefficients` types and then returning them here and then require those in `calculate_transforms`.

This is definitely something we can leave for now, though I wanted to record this somewhere.

sb_vision/tokens.py
PeterJCLaw: Unfortunately we don't have any tests for `pixel_coords`, so it would be good to check manually that these are the same as they were previously.

The test cases I'd check would be:
 * marker in directly front of camera, but _facing_ 45° to one side
 * marker directly in front of the camera leaning backwards by 45°
 * marker flat on to the camera, but rotated 90° and 180° so it's on its side/upside down

If that results in some useful demo images adding them as test cases might be useful, particularly if we also want to expose `orientation`.

I think for this PR we need to do enough to check that we're not changing what `pixel_coords` means; adding the test cases can wait if you like. (Alternatively, if you do want to put that into this PR, maybe branch this PR to another which targets it and put the change there so it's easy to review in isolation).

sb_vision/coordinates.py
```diff
@@ -51,7 +51,7 @@ def cartesian_to_legacy_polar(cartesian: Cartesian) -> LegacyPolar:
     """
     Convert cartesian co-ordinate space to the legacy "polar" space.
 
-    This is kept for compatibilty only.
+    This is kept for compatibility only.
```
PeterJCLaw, line 57: Please cherry-pick this to `master`.

sb_vision/tokens.py
PeterJCLaw: I don't think we should allow this argument to be omitted.

PeterJCLaw (dismissed):
@Adimote thanks for taking this on; this looks like it's going to be a much more reliable solution!

I've had a good look through the code, though haven't had a chance to run this on a Pi. I'd like us to do that soon.

I've raised quite a lot of points here, many are small or stylistic (I realise we don't have an explicit style guide, though I do have a strong sense of the style of this repo so I've tried to preserve that; please excuse me acting a bit like a linter here), though some are a bit larger. I've tried to include a lot of detail where I can and highlighted when things are longer term ideas rather than must fix.

I've also highlighted some redundant changes and code it would be good to remove so as to minimise the changes being introduce here; for simplicity of understanding the changes both by us now and anyone in the future.

To collect them into one place, the big things I think we need to do are:
- working out how we package this given that it adds dependencies
- testing this on a Pi (if hasn't been done already)
- document how to calibrate a camera in this new mechanism; this probably needs a short paragraph in the README plus a link to the OpenCV docs

Most of the other changes are simpler things we can just fix.

sb_vision/distance_finding.py
PeterJCLaw: Turns out mypy will only complain about partial signatures when `disallow_untyped_defs` is set to `True`. I guess that makes sense, though it would be nice if there was a `disallow_partial_defs`.

sb_vision/coordinates.py
Adimote: fixed

sb_vision/distance_finding.py
Adimote: the return value, (true / false) which I assume will only be false if there are incorrect parameters. It seems all functions return a `retval`, some just return 1, others return a double.

tests/test_coordinates_from_file.py
Adimote: fixed in b6131de8487f164f1eeb1b642844ed16b8c8b8dc

sb_vision/distance_finding.py
Adimote: Fixed in f21d2628d1fc6f671e82541fdf94f27589cf77bd

Adimote: Fixed in f21d2628d1fc6f671e82541fdf94f27589cf77bd

sb_vision/distance_finding.py
Adimote: fxed in 65503146f0f7a4255f7ea5600ce26807da5ffa62

Adimote: fixed in 65503146f0f7a4255f7ea5600ce26807da5ffa62

sb_vision/tokens.py
Adimote: Previously it was only present when a distance model was passed in.

The equality check for a token only ever uses the `id`, so I would argue the minimum needed for a valid token should only be the `id`. Also there are tests which instantiate a token with a blank id currently which I think are fine.

This means that a token can be made in 3 ways, it can be just an id, an id with pixel co-ordinates (when apriltag runs), or an id with pixel co-ordinates and 3D co-ordinates (when a camera model is passed in). I think this makes sense. 

Forcing the token to have `pixel_coords` binds it to an image when it otherwise isn't.

setup.py
Adimote: I think you're under the assumption that our current system works better than it does. It's actually practically unusable right now. I really don't think there's any point in supporting both. IMO the only option if we can't deploy `opencv-python`  is to write our own OpenCV bindings again. In which case it won't be a revert, it will be an addition of new openCV bindings

sb_vision/distance_finding.py
Adimote: fixed in 97b6f78d682ba991a7b762ca6f0ac094cb33f1c2

sb_vision/distance_finding.py
Adimote: fixed in 0ecfb436d2927909e9ac6fe6c6b34150e22a9350

sb_vision/distance_finding.py
Adimote: fixed in 15e2ce14227ee238a7f029c7a5f9a8b165db2ee3

sb_vision/distance_finding.py
Adimote: fixed in a50c9cbee486944a2437d519030bff65cb1eebe9

sb_vision/distance_finding.py
Adimote: fixed in a50c9cbee486944a2437d519030bff65cb1eebe9

sb_vision/distance_finding.py
Adimote: fixed in a50c9cbee486944a2437d519030bff65cb1eebe9

sb_vision/distance_finding.py
Adimote: a50c9cbee486944a2437d519030bff65cb1eebe9

sb_vision/distance_finding.py
Adimote: removed in a50c9cbee486944a2437d519030bff65cb1eebe9

sb_vision/distance_finding.py
Adimote: fixed

tests/test_coordinates_from_file.py
Adimote: fixed in edfc7bcb6d5c9589c62c358d9506e155f3082579

tests/test_coordinates_from_file.py
Adimote: fixed in edfc7bcb6d5c9589c62c358d9506e155f3082579

tests/test_coordinates_from_file.py
Adimote: I don't really like functools.partial as I think it isn't as readable as defining a lambda.

I've come up with what I think is a nice solution here:
daf8dc949bb196d71f5f9d1b70f4770008154646

tests/test_marker_sizes.py
Adimote: fixed in e9a01eca6ca8e4dbfee04ad9e33ff60d9686566c

tests/test_coordinates_from_file.py
Adimote: fixed in 071ac22a07db3dedc4482ac910662f8d2913df58

tests/test_coordinates_from_file.py
Adimote: fixed in 071ac22a07db3dedc4482ac910662f8d2913df58

sb_vision/distance_finding.py
Adimote: 734bd0c5ea7e58cfd14e93b0c355e2c03f0b46d5

sb_vision/distance_finding.py
Adimote: 30cc29648839d7db5a8cbe287a4d37eabe6cbc27

sb_vision/distance_finding.py
Adimote: Because you'll probably ask for it, I've added an excpetion upon a false value at 8527bd46a73e80d744fd5d6ed44e1ab7a7e43073

sb_vision/distance_finding.py
Adimote: For some reason solvePnP returns a vertical array (a list of lists, each having one item), so we're not throwing anything away here.

sb_vision/tokens.py
Adimote: I think I've got a good name for it: 'update_3D_transforms' 4a494b73e5f8fb932d1048344b8326ebcf3370e0

sb_vision/tokens.py
Adimote: Yes, it's going to be added in the next PR

Adimote: I've removed it for now in 5ca48ba97d3719fe10a4de248254e739f2e8b8bb

sb_vision/tokens.py
Adimote: 9ce4baa58cca20362e91f3df8e30a3217f05600c

sb_vision/tokens.py
Adimote: Fixed in 0c947e29d56f76eaffe89d46205ec1f345a76cd3

Adimote: I would argue Cartesian describes a translation of any point from 0, and thus Cartesian is suitable for both values. It allows for future extensions to be more useful (i.e. adding translations together will be exactly the same as adding Cartesian co-ordinates together)

sb_vision/vision.py
Adimote: You're correct, fixed in 7d28b98bbebed9988468f6ec43bb101b37500ee2

sb_vision/tokens.py
Adimote: Good catch! turns out the order of the corners aren't what were previously. IMO our system actually makes less sense than the ones which Apriltags returns, so I'll convert them last minute to the marginally less useful version to return to robot-api. 3c8f8812c0fa39780c752ad628f0458208f57ce3

sb_vision/distance_finding.py
PeterJCLaw: On the one hand: ewww.
On the other hand: there must be some complicated reason it does that.

sb_vision/distance_finding.py
```diff
@@ -8,25 +8,26 @@
...
 
 def _get_values_from_xml_element(element: etree.Element) -> List[str]:
     """Parse an xml tag with space-separated variables."""
     text = []  # type: List[str]
     for e in element.itertext():
-        if e.strip():
-            text += re.split('\s+', re.sub('\n', '', e.strip()))
+        e = e.strip()
+        if e:
+            text += re.split(r'\s+', e.replace('\n', ''))
```
PeterJCLaw: I've just realised that there might be an edge-case bug here: if `e` contains `'foo\nbar'` then this will extend `text` with `['foobar']`, when I think it should instead extend it with `['foo', 'bar']`. I'm familiar with the spec of the file, so don't know whether that's expected.

If it's not expected (and the newline should count as whitespace), then we can just drop the `.replace('\n', '')` entirely as the newline will be matched as whitespace.

Side note: I was actually expecting to need to pass an extra flag to allow the newline to count as whitespace, but it appears that that's a default:
``` python
>>> re.split(r'\s+', 'aa\nbb')
['aa', 'bb']
```

sb_vision/distance_finding.py
```diff
@@ -39,60 +40,63 @@ def get_calibration(file_name: Path) -> Dict[str, Any]:
...
-            if 'type_id' in element.attrib and element.attrib[
-               'type_id'] == 'opencv-matrix':
-                rows, cols = int(element.find('rows').text), int(
-                    element.find('cols').text)
-                data_type = element.find('dt').text
-                values = _get_values_from_xml_element(element.find('data'))
-                if data_type == 'd':  # doubles
-                    data = np.reshape([float(v) for v in values],
-                                      (rows, cols)).tolist()
+            if element.tag in ['dist_coeffs', 'cameraMatrix']:
```
PeterJCLaw: Minor: it's occurred to me that, if `lxml` supports it, searching for these elements by name and then running the conversion of their contents on the result might be a slightly nicer way to spell this than the loop and conditionals that we have here. It would also mean we could generate a nicer error if the calibration file is missing either of them.

Assuming the body here was extracted to a util, that might look something like this:
``` python
    def convert(element): # probably a better name
        ... # implementation pretty much as current body

    dist_coeffs = convert(tree.find('dist_coeffs'))
    camera_matrix = convert(tree.find('cameraMatrix'))
    return dist_coeffs, camera_matrix
```

The current code appears to be functional, so we can probably defer this if desired.

sb_vision/distance_finding.py
PeterJCLaw: Please can you address this or explain why not?

sb_vision/distance_finding.py
```diff
@@ -109,13 +113,26 @@ def calculate_transforms(
...
+        np.array(pixel_corners),
         np.array([np.array(xi) for xi in camera_matrix]),
         np.array([np.array(xi) for xi in distance_coefficients]),
     )
-    translation_vector = tuple(v[0] for v in translation_vector)
-    orientation_vector = tuple(v[0] for v in orientation_vector)
+    if not return_value:
+        raise ValueError("cv2.solvePnP returned false".format(return_value))
+
+    translation_vector = Cartesian(*(v[0] for v in translation_vector))
```
PeterJCLaw: I appreciate the desire to use the types here, though I'm not sure it's a good idea to have values in a named tuple that don't match their intent.

sb_vision/tokens.py
PeterJCLaw: Please can you address the comment about the typing of `pixel_centre` or explain why not?

sb_vision/tokens.py
```diff
@@ -51,57 +50,57 @@ def from_apriltag_detection(
...
         marker_id = apriltag_detection.id
 
         # *************************************************************************
         # NOTE: IF YOU CHANGE THIS, THEN CHANGE ROBOT-API camera.py
         # *************************************************************************
+
+        # The pixel_corners value we expose has the first one in the bottom left.
+        # Apriltags returns the first one in the top left. thus we need to rotate
+        # the values by one to get the correct ordering.
+        offset_pixel_corners = [pixel_corners[3]] + pixel_corners[:3]
```
PeterJCLaw: Can we clarify whether this the top left of the _marker_ (i.e: relative to what the marker defines as the "top"), or the top left of the _image_ of the marker (i.e: relative to the image itself). (Or something else)

PeterJCLaw: It turns out that we don't appear to have documented the presence of the `pixel_corners` attribute, so I'm a bit happier about changing it if we felt we needed to do so.
Obviously we still need to understand what they mean and then document that either way.

sb_vision/tokens.py
PeterJCLaw: Unfortunately I never really liked that one could construct a `Token` just with an `id`, nor that the equality was based purely on the ids. From what I can tell, that was that was there only to simplify testing and is something I was hoping to address at some point.

Anyway, this change is actually breaking -- previously a `Token` without `pixel_coords` or `pixel_centre` completely didn't have those attributes. Now those attributes are always present and sometimes have empty values.

See [tokens.py:213-231](https://github.com/sourcebots/sb-vision/tree/463f08731e4ce6b9cd7033f254fa43c43485e12e/sb_vision/tokens.py#L213-L231) for the implementation -- the attributes are only defined via `infer_location_from_homography_matrix`, though that is always called from `from_apriltag_detection` and always sets the attributes _before_ checking whether a model was provided.

PeterJCLaw:
@Adimote did you know you can ask for re-review without dismissing the previous review?

PeterJCLaw (dismissed):
Looking pretty good now. Request changes as I think I've spotted a bug, but we're nearly there.

There were a couple of review comments from the previous round which appear to have been overlooked. I'm not sure if this was just because they got lost or if there's some reason they can't be fixed.

RealOrangeOne:
As they're no longer required, could we delete the tecknet model files in this PR too.

setup.py
RealOrangeOne: I agree with @PeterJCLaw here. If we can't actually deploy this PR, there's no use it going into master. 

If there are no packages available for `opencv-python`, then we _could_ create one (or prompt upstream to do so), it doesn't look especially difficult, although would lead to some changes to `pi-image` obviously. Alternatively, I don't think there will be much harm in writing our own bindings, seeing as we're using a very minimal set of the `opencv` functions (and i can't see packaging `opencv` being an issue!).

sb_vision/tokens.py
Adimote: Fixed in 7ef18e11726e05212cd6daff2e542f289701665b

Adimote: Fixed in 7ef18e11726e05212cd6daff2e542f289701665b

sb_vision/distance_finding.py
Adimote: fixed in 96167c6215cdc3d493faa10e886955cbf51e8d17

sb_vision/distance_finding.py
Adimote: I really like this change, have put it in cda226dee2383bfb0518257ccbaf1b898b6c8b78

sb_vision/distance_finding.py
Adimote: Apologies, looks like some comments fell through the gaps, fixed in c17740db70244cc64dd15010f846f93c125043bf

sb_vision/tokens.py
Adimote: My OCD is killing me that this exists, however a team mentioned to me that they found this function, assuming they're using it, we can't silently change its behaviour just to remove one line of code

Adimote: I've carded up the task to undo this at https://trello.com/c/w9BzXPUV/81-change-pixelcorners-behaviour-to-be-the-same-as-apriltags so we can restore normality after this years game

Adimote: I think I've clarified it in a simple way in 40688997f88f59c26beb3a7689e3253f2f399149

Adimote:
@RealOrangeOne I've fixed your requested change in 227508fa85410d1c6a796ad2c2c4d0186da6a401

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,139 @@
...
+    ])
+
+    return_value, orientation_vector, translation_vector = cv2.solvePnP(
+        object_points,
+        np.array(pixel_corners),
+        np.array(camera_matrix),
+        np.array(distance_coefficients),
+    )
+    if not return_value:
+        raise ValueError("cv2.solvePnP returned false".format(return_value))
```
RealOrangeOne: I think you're missing a format placeholder in this string?

sb_vision/distance_finding.py
```diff
@@ -0,0 +1,139 @@
+"""
```
RealOrangeOne: I'm not sure I like the name of this file. Whilst this file does partially handle finding the distance, it also handles _lots_ of other things.

Adimote: That's fair, fixed in d40ccf21b70b466a98e664398d17ed8feb4db40c

sb_vision/distance_finding.py
Adimote: Good catch, fixed in 79098ff764ec9508bd6431282c8cd443b07d2c5e

PeterJCLaw:
So Github doesn't really have a good way of showing this sort of thing. I think the idea would be having some built-in way of saying "this depends on $branch, but will be merged to master once that merges" which would then block the merge (at least in the web UI) until the relevant branch had merged.

For this PR, I think it makes most sense to point it at #33 until just before that merges, then to switch it to master. Unfortunately if you point a PR at another PR and then the target PR (e.g: #33) gets merged, then the second-layer PR (e.g: this one) gets closed.
I can't remember whether you can then re-open it (I've a recollection that the UI doens't like it, since the target branch might not still exist). This also plays particularly badly with the deletion of branches which are now merged (as we tend to do, and seems reasonable on its own).

tests/test_coordinates_from_file.py
Adimote: Yes, they're actually as detailed as not using approx

sb_vision/tokens.py
Adimote: Fixed a while ago (CBA to find commit number)

sb_vision/distance_finding.py
Adimote: Fixed a while ago (CBA to find commit hash)

sb_vision/find_3D_coords.py
```diff
@@ -0,0 +1,139 @@
...
+    return_value, orientation_vector, translation_vector = cv2.solvePnP(
+        object_points,
+        np.array(pixel_corners),
+        np.array(camera_matrix),
+        np.array(distance_coefficients),
+    )
+    if not return_value:
+        raise ValueError("cv2.solvePnP returned false")
+
+    translation_vector = Cartesian(*(v[0] for v in translation_vector))
```
PeterJCLaw: I don't think it's a good idea to put values into this typed named tuple which don't match the contract of the type.

sb_vision/find_3D_coords.py
```diff
@@ -0,0 +1,139 @@
...
+    """Parse an xml tag with space-separated variables."""
+    text = []  # type: List[str]
+    for e in element.itertext():
+        e = e.strip()
+        if e:
+            text += re.split(r'\s+', e)
+    return text
+
+
+def convert(element: etree.Element) -> List[List[np.float64]]:
```
PeterJCLaw: Is there really not a better name for this? (Also: should it be private, i.e: `_`-prefixed?)

sb_vision/find_3D_coords.py
```diff
@@ -0,0 +1,139 @@
...
+    """Converts an element containing an OpenCV matrix to python lists."""
+    type_id = element.attrib.get('type_id')
+    if type_id == 'opencv-matrix':
```
PeterJCLaw: Style question: I note that you tend to prefer code along the lines of:
``` python
if condition:
    ...
    return ...
else:
    raise ...
```
rather than:
``` python
if condition:
    raise ...
...
return ...
```
Is there a reason for that?

My preference is for the latter as I find it easier to read & parse (i,e: it has less cognitive load) and like that it results in less nesting (which I find helps make it easy to understand).

sb_vision/tokens.py
```diff
@@ -13,164 +9,28 @@
...
 class Token:
     """Representation of the detection of one token."""
 
-    def __init__(self, id: int, certainty: float=0.0) -> None:
+    def __init__(
+        self,
+        id: int,
+        certainty=0,
+        pixel_corners: List[PixelCoordinate] = None,
+        pixel_centre: PixelCoordinate = PixelCoordinate(0.0, 0.0),
```
PeterJCLaw: Allowing these arguments (`pixel_corners` and `pixel_centre`) to be optional with empty values when omitted is breaking -- previously they would always have valid values if present.

(I realise there was a previous thread on this, but it seems to have got lost in the file rename)

PeterJCLaw (dismissed):
Mostly good now.

I still think we shouldn't change the behaviour of `pixel_corners` and `pixel_centre` though.

See other smaller comments inline.

sb_vision/find_3D_coords.py
Adimote: I don't really understand what you mean here, the type is a triplet of AnyFloat, and I'm passing in a triplet of floats, in what way does it not match the contract?

sb_vision/find_3D_coords.py
Adimote: fixed in fcbd21e77acd6c677d8b24fde4921beffe29d871

sb_vision/find_3D_coords.py
Adimote: I think this is a symptom of how I write logic. I prefer to jump into the core logic and work on how to handle edge cases afterwards. If I think of an edge case while writing my code, I usually just stick in an `if` statement to filter away the badness and continue writing. Once the core logic is done, I then think about how to handle edge cases, and the minimal change from there is to just add an `else` with the code to handle them.

I understand the readability benefits of the latter method, so I think I'll try to get into the habit of moving code to the second method once it's written.

Adimote: Another thing is it makes it easier to handle if I wanted to later expand the code to handle more than one condition. In this case I wrote that function with the intention to make it easy to expand to more complex files, hence why in the first version of this, I initially put everything in a dictionary and returned that.

Adimote: Anyways, I've fixed this in faa74bd5333d5bea6872e1eb36b7211668d4a8c7

sb_vision/tokens.py
Adimote: I'm not sure what you mean by 'breaking' here, is there somewhere else in the code where there's a check for `pixel_corners` / `pixel_centre`?

Either way I've moved the pixel specific information into a separate function in 695d65576f59401818b83b3e92b3a269738fc207 and made them undefined if not set.

sb_vision/find_3D_coords.py
PeterJCLaw: The contract is not just that it's a type -- we also expect that positive y means up. However the conversion which follows indicates that that wasn't actually the case here. I realise that that's not actually documented on the type itself, though the README does specify this.

This actually goes away in my `solvePnP` wrapper (#36) as I've pushed this fix down into the C layer.

PeterJCLaw:
@Adimote thanks for those changes; I think this branch is now all good except the deployment plan & testing it on a Pi.

My wrapping of `solvePnP` now works and I've put up #36 with code.

As our usage of `lxml` was fairly lightweight, I had a quick look at whether we could remove that. It turns out that the `etree` stuff we were using is present in the standard library. I'm not sure whether it's actually the same code, but it works the same as far as we were using it. I've pushed #37 which moves us to the standard library there.

The combination of those allows us to ship this without any extra packages, which makes this much easier to ship! Once those PRs are merged into this one, this PR can be merged into master.

PeterJCLaw:
@Adimote I've just spotted that this still lists some of the now-redundant dependencies in the debian package control file. We should remove those.

Specifically we can remove `python3-sklearn` and `python3-yaml`.

PeterJCLaw:
@Adimote manual testing of this on a Pi mostly works, huzzah!

There are a couple of easily fixable breakages:
- `robotd` will currently always request a calibration for a `'c270'` model. We therefore need to not explode if that's asked for. We can do this either by faking the c270 calibration as being the C016 calibration (probably via a symlink) or by actually calibrating a c270. My preference would be for the latter as it's less confusing for anyone reading this repo and will allow developers to use them if that's what they have to hand.
  (I worked around this by editing my `robotd/camera.py` to ask for `'C016'`)
- `robotd` expects that a `Token` will have a `homography_matrix` attribute (with a `.tolist()` method). We'll need to continue to expose that so as not to break `robotd` users (I don't see any reason to remove it really).
  (I worked around this by commenting out that line in my `robotd/camera.py` )

Other than that, this now works well and seems to be fairly accurate :)

PeterJCLaw (approved):
:shipit: Woo!

mxbi:
Now that this has been merged, is the new vision system in a state where we can load it up on the pi (after installing the libopencv/python headers) and try it out? :)

PeterJCLaw:
@mxbi happily there aren't any new dependencies for this change, so all that's needed is to build the latest package and install it.

Having done that you should find that the existing support in `robot-api` yields much more accurate results, though you may wish to wait for the next published update as that will include some other niceties too.

PR #36, opened 2018-02-25
Wrap solvePnP directly

PeterJCLaw:
This introduces a new cffi module which wraps the [`solvePnP`](https://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html#solvepnp) function we're now using for 3D transforms.

### DIFF
--- a/sb_vision/cv3d.py
+++ b/sb_vision/cv3d.py
@@ -31,7 +31,7 @@ def _ffi_flattened_float_array(values: Sequence[Sequence[float]]):
     flattened_values = sum([tuple(x) for x in values], ())
     count = len(values) * lengths[0]
 
-    return _cv3d.ffi.new('float[{}]'.format(count), flattened_values)
+    return _cv3d.ffi.new('double[{}]'.format(count), flattened_values)
 
 
 def solve_pnp(
--- a/sb_vision/cv3d.py
+++ b/sb_vision/cv3d.py
@@ -1,0 +1,72 @@
+"""
+Python wrapping around cffi layer to interop to OpenCV 3D functions.
+"""
+
+from typing import Sequence, Tuple
+
+from sb_vision.native import _cv3d  # type: ignore
+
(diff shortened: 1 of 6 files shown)


sb_vision/cv3d.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def _ffi_flattened_float_array(values: Sequence[Sequence[float]]):
+    if not values:
+        raise ValueError("Refusing to create ffi matrix for empty 'values'.")
+
+    lengths = list({len(x) for x in values})
+    if len(lengths) != 1:
+        raise ValueError(
+            "'values' is not rectangular (got row lengths of {} and {})".format(
```
Adimote, line 25: very very minor: this doesn't produce an error message with an oxford comma :smile:

sb_vision/native/cv3d.cpp
```diff
@@ -0,0 +1,100 @@
...
+int solve_pnp(
+    const double object_points[],
+    const double image_points[],
+    const double camera_matrix[],
+    const double dist_coeffs[],
+    double rvec[],
+    double tvec[]
+) {
+    // Copy our input data into OpenCV's Mat wrapper
+    cv::Mat object_points_mat(4, 3, CV_64FC1);
```
Adimote: I've seen other implementations of this use `cv::DataType<double>::type` instead of `CV_64FC1`, I'm not sure if the former is nicer though.

Adimote (approved):
Seems like a very nice implementation. Two comments but they're only nitpicks, really. (One is mostly a joke)

sb_vision/native/cv3d.cpp
PeterJCLaw: Interesting, I think that that's probably nicer as it's clearer what the constant means.

PeterJCLaw: 5d32563 does this; thanks! (I'd not realised that that existed)

PR #40, opened 2018-02-25
Produce an error when loading a mismatching calibration

PeterJCLaw:
If we use a calibration for a resolution other than that which it is intended, we'll get nonsense values. Ensure that that doesn't happen by requiring that `load_camera_calibrations` is passed the image resolution which will be used and then validating that against the value in the calibration file.


tests/test_load_calibrations.py
```diff
@@ -0,0 +1,32 @@
...
+    with mock.patch(
+        'pathlib.Path.open',
+        return_value=(TEST_DATA / 'C016_calibration.xml').open(),
+    ):
+        with pytest.raises(ResolutionMismatchError):
+            load_camera_calibrations('camera-model', (100, 100))
+
+
+def test_accepts_matching_resolution():
+    """Ensure that loading a mismatched calibration file produces an error."""
```
Adimote: Looks like you forgot to update this docstring

Adimote (changes requested):
Looks excellent as usual, there's an incorrect docstring however.

PeterJCLaw:
Apologies, I should have held fire before PRifying this; rebase incoming.

tests/test_load_calibrations.py
```diff
@@ -0,0 +1,51 @@
...
+    with mock.patch(
+        'pathlib.Path.open',
+        return_value=(TEST_DATA / 'C016_calibration.xml').open(),
+    ):
+        with pytest.raises(ResolutionMismatchError):
+            load_camera_calibrations('camera-model', (100, 100))
+
+
+def test_accepts_matching_resolution():
+    """Ensure that loading a matcing calibration file works."""
```
Adimote: Miss-spelt `matcing`

tests/test_load_calibrations.py
PeterJCLaw: Yup; copy/pasta error: bd5fd55b26efc21b16f84ead192b85589b5cc37c

tests/test_load_calibrations.py
PeterJCLaw: hnnng

PeterJCLaw: dada343

Adimote (approved):

Adimote:
:shipit:

PR #42, opened 2018-03-07
Bump version to 2018.2.0

Adimote:


RealOrangeOne (dismissed):
Missing bump to `setup.py`

debian/changelog
```diff
@@ -1,4 +1,4 @@
-sb-vision (0) unstable; urgency=low
+sb-vision (2018.2.0) unstable; urgency=low
```
PeterJCLaw: Are we going to do this to all the other packages too? I've no idea whether they depend on specific versions.

This **needs** testing before this merging.

PeterJCLaw: I'm pretty sure we should _add_ to the entries in this file, not replace them.

PeterJCLaw (dismissed):

debian/changelog
Adimote: I've updated and tested this :+1:

Adimote: Good point

PeterJCLaw: For our own future reference, could you detail somewhere what the testing procedure was?

Adimote: I've documented this in https://github.com/sourcebots/wiki/wiki/Update-file-Testing-Process

Adimote: Fixed in aba4106

RealOrangeOne (approved):

PR #44, opened 2018-03-21
remove the 'if changed then change there' comment

Adimote:
Remove the shouty comment as there actually isn't any tight coupling between `sb_vision` and `robot-api`, they should both be updated but it's no different to changing the spec of an API.


PeterJCLaw (approved):
Minor meta point: I'd be happy for this sort of thing (trivial, already discussed) to just be pushed to master. It's not really worth the round-trip-with-interruptions time that it takes to get review for this sort of thing.

PR #45, opened 2018-06-04
Rearrange pixel corner co-ordinates

Adimote:
Rearrange the marker corners to match that of apriltags

### DIFF
--- a/sb_vision/tokens.py
+++ b/sb_vision/tokens.py
@@ -89,7 +89,7 @@ class Token:
 
         self.homography_matrix = homography_matrix
 
-        # pixel corners: clockwise from the top left
+        # pixel corners: clockwise from the top left corner (relative to the marker) of the marker
         self.pixel_corners = pixel_corners
 
         # centre of marker: average the corners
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
@Adimote Are these relative to the marker or the image?

That is, if the marker is upside down: is the first position value here that of the upper-leftmost corner within the image or the bottom-rightmost corner in the image?

Adimote:
@PeterJCLaw yes that's correct

Adimote:
@PeterJCLaw I've made it more clear in the comment

PeterJCLaw (approved):
Even better would be to include an example explaining what it means for the coordinates to be relative to the marker.

Adimote:
Turns out I was completely incorrect about the ordering of the corners. That shows me for trusting my memory. I've now updated the ordering to match actual values

PeterJCLaw (approved):
Given the level of uncertainty around this, maybe a test to demonstrate example values is in order?

PeterJCLaw:
Speaking of testing -- the CI is highlighting some long lines.

PR #46, opened 2018-06-24
Remove Legacy Polar Co-ordinates system

trickeydan:
Do not merge until this system has been removed from both `docs` and `robot-api`

### DIFF
--- a/sb_vision/tokens.py
+++ b/sb_vision/tokens.py
@@ -2,6 +2,8 @@
 
 from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
+import numpy as np
+
 from .coordinates import Cartesian, cartesian_to_spherical
 from .find_3D_coords import (
     PixelCoordinate,
--- a/sb_vision/tokens.py
+++ b/sb_vision/tokens.py
@@ -2,12 +2,7 @@
 
 from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
-import numpy as np
-
-from .coordinates import (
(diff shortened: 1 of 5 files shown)


trickeydan:
@Adimote There's a strange failure with the tests here. The linting complains that `np` is unused in 38baa8d and 741d705, yet when I removed it in a89f0bf a load of errors occur due to `np` not being imported.

sb_vision/tokens.py
```diff
@@ -110,9 +106,6 @@ def update_3D_transforms(
         self.cartesian = translation
 
         # Polar co-ordinates in the 3D world, relative to the camera
-        self.polar = cartesian_to_legacy_polar(self.cartesian)
-        self.legacy_polar = cartesian_to_legacy_polar(self.cartesian)
```
PeterJCLaw, line 114: We should ensure that the `robotd` changes to remove `legacy_polar` happen _before_ this change is merged.

PeterJCLaw (commented):
Seems good; I'll have a more complete look when I've time to test it out.

We should check that `robotd` (and `robot-api`) have stopped using these features before removing them. I know we want to do that, so hopefully this is just a case of remove the dead code there too (first).

CircleCI is highlighting that there's an import of `numpy` in `coordinates.py` which is unused. Note that `coordinates.py` imports `numpy` both as a module and some functions from it; only one of these can be removed.
(It looks like the CI previously correctly highlighted the erroneous removal of a similar import from `tokens.py`)

